Issue #7, opened 2022-01-05
High latency when audio is enabled.


danisla:
Enabling audio results in high latency (700-1000ms) after first connection. It eventually drops down to something more reasonable (100-250ms).

ehfd:
xpra-html5, a similar project with a different direction, seems to want to decode audio in a [worker](https://github.com/Xpra-org/xpra-html5/issues/89) or use [Web Audio API](https://github.com/Xpra-org/xpra-html5/issues/64).

> Meh. Probably best to use Web Audio API (https://github.com/Xpra-org/xpra-html5/issues/64) instead.
FWIW, this fork for aurora is more up to date: https://github.com/Kukunin/aurora.js - https://github.com/Kukunin/opus.js

More technologies available to explore: https://github.com/Xpra-org/xpra-html5/issues/56

danisla:
The xpra-html5 project has very high latency for audio (10-45 seconds). Not sure if this is because of an outdated opus decoder or just the nature of the websocket and latency.

If we want to keep the audio in sync with the video, the gstreamer pipeline has to be driven by the clock of the audio stream.

If it's just an issue with how PulseAudio is being integrated with GStreamer, maybe there is some bug, or we look into using [PipeWire with PulseAudio](https://wiki.archlinux.org/title/PipeWire) to see if we can reduce the latency.

I was also thinking about offering a decoupled audio/video experience, where we send the audio in a separate stream from the video, sacrificing audio synchronization. This could be done with a new webrtc transceiver and leveraging most/all of the browsers MediaSource capabilities for decoding and buffering. Another option would be to send opus packets over a data channel and decode them with a WASM OPUS decoder like what Aurora is doing.

The bigger issue with MediaSource in the browser is that it's designed for voice applications (video conferencing) so there are algorithms optimized to maintain audio continuity and synchronization with latency as the tradeoff. Unfortunately, these algorithms and their jitter buffers cannot be configured.

totaam:
> If we want to keep the audio in sync with the video, the gstreamer pipeline has to be driven by the clock of the audio stream.

Audio cannot have *any* jitter or it sounds terrible so the receiver always maintains a buffer. It's the video that must be synced with the audio.
The xpra server detects video content and then delays video frames (and only the video part of the window, nothing else) by the same amount.

> If it's just an issue with how PulseAudio is being integrated with GStreamer ...

Unlikely. The server-side latency is very small.

[...]

PR #31, opened 2022-04-19
Update to GStreamer v1.20

danisla:
Notable changes:

- Build gstreamer 1.20 from monorepo
- Include gstwebrtcbin patch to add `rtx-time` to SDP per [#7 comment](https://github.com/selkies-project/selkies-gstreamer/issues/7#issuecomment-1100765217).
- Enable `do-nack` on transceiver to reduce stuttering/freezing on packet loss, thanks to @xhejtman.
- Set RTP cap `aggregate-mode` to `zero-latency` to reduce RTP packetization overhead.
- coturn-web now uses "?transport=tcp" param for turn URI to enable traversal of networks with UDP blocked.

New dependencies:
- Need to `apt-get install x264` in target images to use software encoder.

TODO:
- [x] Verify x264enc encoder option still works.
- [x] Verify nvh264enc encoder option still works.


xhejtman:
```
rtph264pay_caps.set_value("aggregate-mode", "zero-latency")
rtph264pay_caps.set_value("rtcp-fb-ccm-fir", True)
```

perhaps set these two as well

ehfd:
Feedback: @danisla @xhejtman 
The screen freeze of multiple seconds in v1.2.0 and before has completely disappeared.
However, the screen freeze seems to have changed in a fashion where there the screen freezes for fractions of seconds more frequently. It's better than before, but we still might be able to improve a bit.

xhejtman:
are you seeing significant number of packet loss? my observation is that 10% of packet loss is way too much for smooth experience. does it change with bitrate? and are you seeing that nacks/pli is actually sent? this is available in chrome://webrtc-internals in statistics. Also you should see that only 2-3 key frames are sent in total.

ehfd:
Packets are not getting lost. But the screen is still tearing and briefly pauses for 0.x seconds every 4-5 seconds.
image

ehfd:
@xhejtman Might be my internet, let me update my router firmware and try again...

ehfd:
@xhejtman Feels much better after improving my internet (resetting my router).
Latency is exactly 2 times the ping to the server as well, unlike before, where 10-20 ms of extra latency was added.
It's smooth, with only very occasional tearing which doesn't impact the overall experience.

xhejtman:
well, we cannot do much if connection sucks ;) perhaps we can optionaly enable FEC? But latency will suffer and I am not sure it actually helps a lot. But Geforce Now stream offers also FEC enabled stream. I wonder if and how we can offer several stream options?

PR #32, opened 2022-04-19
Fix input exception when browser window is unmapped

xhejtman:
If stream is runing and user e.g., switch desktop or tab for a while, streamer can get exception as mouse variables are nil.

Signed-off-by: Lukas Hejtmanek <xhejtman@gmail.com>


danisla:
Great find and thanks for the fix!

Issue #11, opened 2022-02-03
Client side web interface does not fit correctly in HiDPI clients


ehfd:
![image](https://user-images.githubusercontent.com/8457324/152354030-3ddb59cd-73af-46ab-80e4-b62fa746b3dc.png)
![image](https://user-images.githubusercontent.com/8457324/152354675-0255eda9-c0b7-4718-b3cd-e7d369bd54df.png)

When not used full-screen, local scaling interface does not fit into the window of the web browser in HiDPI clients. Reproducible by setting PC DPI settings above 100% (96 DPI for Linux). 125% (120 DPI for Linux) requires a slight scroll to see the taskbar, and 150% (144 DPI for Linux) requires more scroll to see the taskbar.

Explore how noVNC and Guacamole implements local scaling.
Starting points:
https://github.com/novnc/noVNC/blob/master/app/ui.js
https://issues.apache.org/jira/browse/GUACAMOLE-164
https://issues.apache.org/jira/browse/GUACAMOLE-165

danisla:
Rather than scaling locally, HiDPI is best supported by changing the remote. Doing local scaling would mean the resulting image would be blurry.

From my initial research, this would have to be done per window manager, I've seen this done before with Mutter.
https://wiki.archlinux.org/title/HiDPI

I am not sure if there is a window manager agnostic way to configure this.

ehfd:
I did more investigation, and this specific issue I'm seeing is with the client HTML5.
![image](https://user-images.githubusercontent.com/8457324/152470791-0f7c130b-172a-4ba0-8d1e-8cb6c9fab2ba.png)
Since the scroll is required since before the screen is started.
Likely will need to look at Javascript.

ehfd:
Should be changed from enhancement to bug...

xhejtman:
Do you use some tool bar like bookmarks bar? I believe this is because of tool bars. For some stupid reason, `document.body.offsetHeight` contains height **including** tool bars which breaks all things. I do not know how to workaround this. It seems that `height: 100%` canvas overflow in this case. It's get fixed once you resize chrome window, at least for me.

ehfd:
Hi! @xhejtman 
Thank you for all your comments to this repository.
I develop https://github.com/ehfd/docker-nvidia-egl-desktop and https://github.com/ehfd/docker-nvidia-glx-desktop and have collaborated with this project.
[...]

PR #33, opened 2022-04-21
Fix overflow in UI and fix for Safari

xhejtman:
Should fix overflow in UI if, e.g., bookmarks bar is visible. Also fix for Safari if Permissions api is not enabled.

Please test, if UI issues are gone.


xhejtman:
Should be `vuetify.css.map` regenerated?

ehfd:
Closes #11

danisla:
LGTM, thank you!

PR #59, opened 2022-09-29
[HOTFIX] Fix installation with python-xlib

ehfd:
This addresses two issues.
1. python-xlib breaking in the last commit 10 days ago. Therefore, we stick to the commit where @danisla 's PR was merged to master.
2. pynput wants tkinter and won't run without one, so we add tkinter.


danisla:
LGTM

Issue #40, opened 2022-07-21
[MILESTONE] Make the project more accessible to future contributors and enhance governance


ehfd:
These are the milestones before general availability. I can do some, some I need help.

- [x]  ~~Adequete documentation about setup on bare metal (not just containers), all the way from installing to startup.~~ Done.
- [x]  ~~An activate file that can be sourced to get the GStreamer environment up regardless of which path it is unpacked to~~ Done.
- [x] ~~How to add selkies-gstreamer in a container and also to optionally compile in a container.~~ Done.
- [x]  ~~Make the entrypoint or startup less cumbersome, a quick startup script maybe (with and without an option to start a new Xvfb and also the ability to set which DISPLAY)?~~ Documentation added.
- [x] ~~Include VP8 and VP9 encoding capabilities, although they currently only support software acceleration. The code is already there, but the vpx plugin is not included during GStreamer compilation.~~ Done.
- [x] ~~@danisla Find a way to integrate https://github.com/selkies-project/selkies-gstreamer/issues/6 to Selkies and not leave it to the user/OS or fix the issue altogether.~~ Fixed by #60.
- [x] ~~Check that all of the documentation instructions are valid and none fails.~~ Done.
- [x] ~~Governance event for MPLv2 license.~~ Pull request #62.
- [x] ~~Check fix for https://github.com/python-xlib/python-xlib/issues/233.~~ Fixed by #68.
- [x] ~~Address #43.~~ Patched by #73.
- [x] ~~GStreamer version upgrade. Depends on https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/1605.~~ Upgraded to 1.22.5.
- [x] ~~Regardless, potentially convert docker-nvidia-egl-desktop and docker-nvidia-glx-desktop to selkies-project.~~ Done.
- [x] ~~Because of the warning message `rtpsession gstrtpsession.c:2435:gst_rtp_session_chain_send_rtp_common:<rtpsession0> Can't determine running time for this packet without knowing configured latency`, add a message handler after it does Gst.Pipeline.new(). Something that's going to regularly pop messages from the bus and process them, or a GLib mainloop and a bus signal handler like in that [helloworld example](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-python/examples/helloworld.py).~~
- [x] ~~Update outdated GitHub Actions pipeline.~~
- [x] ~~Pull Request and Issues Template.~~
[...]

PR #60, opened 2022-10-04
Various changes for usability

### LINKED ISSUE #6: Keyboard keys < and > do not work
Pressing either `<` or `>` results in `>` being pressed on the remote.

This breaks tasks like writing HTML in the session.

May need to re-write the python keyboard input handler to use the Xlib directly rather than relying on pynput to do the right thing.
[Example](https://stackoverflow.com/a/18062387)

ehfd:
- Remove tkinter dependency, not needed after all
- Allow empty username for basic authentication if there is no $USER
- Fix docs and base example image behavior
- Ensure that libvpx is explicitly installed and supported
- Change default from nvh264enc to x264enc so the project will start up by default without a NVIDIA GPU
- Remove CPU thread restriction with vpx and x264, encoder will automatically determine number of threads (it was better, at least for me)
- Fix #6.

### DIFF
--- a/Dockerfile.example
+++ b/Dockerfile.example
@@ -98,7 +98,7 @@ RUN \
 
 # Install selkies-gstreamer Python app
 ARG PYPI_PACKAGE=selkies_gstreamer
-ARG PACKAGE_VERSION=1.0.0.dev0
+ARG PACKAGE_VERSION=0.0.0.dev0
 COPY --from=selkies-build /opt/pypi/dist/${PYPI_PACKAGE}-${PACKAGE_VERSION}-py3-none-any.whl .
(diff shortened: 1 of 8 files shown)


ehfd:
image

#6 fixed with minimal side effects.

ehfd:
Closes #6.

ehfd:
Note that we would want GStreamer 1.21.1 before a release when this is merged.

ehfd:
Ready for review @danisla

ehfd:
Fixes a lot of #40 but we'll put the rest in another PR.

ehfd:
@danisla This can be merged. If I have something else I will raise another issue.

danisla:
LGTM

PR #80, opened 2023-04-07
Fix python 3.11+ support

Justsnoopy30:
`sysconfig.get_config_var("SO")` was used in the python-uinput package but the `SO` variable has been deprecated for a long time and no longer works in python 3.11+. This PR updates the dependency to use a fork that changes the variable to its replacement, `EXT_SUFFIX`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ install_requires =
     websockets
     basicauth
     gputil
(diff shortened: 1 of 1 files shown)


danisla:
LGTM. Thanks

danisla (approved):

PR #94, opened 2023-09-13
Move pipeline bus call handler to coroutine

### LINKED ISSUE #93: Gstreamer pipeline is not starting
Gstreamer pipeline is not starting because in __main__.py   lambda preventing to starting pipeline ```loop.run_in_executor(None, lambda: app.start_pipeline()) ``` 
image
after deleting lambda pipeline started 
image

and there is another issue If we change from 

        while self.bus is not None and self.pipeline is not None:
            while self.bus.have_pending():
                msg = self.bus.pop()
                if not self.bus_call(msg):
                    self.bus = None
                    break
            time.sleep(0.1)
            
 to print("started") we see 
image
and that means time.sleep(0.1) triggering infinite sleep and preventing to start the session
If we change time.sleep(0.1) to "break" everything is ok.

danisla:
Allows pipeline to run on 2-core machines.

Fixes #93

### DIFF
--- a/.github/actions/build_and_publish_image/action.yaml
+++ b/.github/actions/build_and_publish_image/action.yaml
@@ -36,6 +36,10 @@ inputs:
     default: "false"
     description: Use cached layers from pre-published image with version 1 (true/false)
     required: false
+  push_image:
+    default: "true"
+    description: Push image to registry
+    required: false
 
 runs:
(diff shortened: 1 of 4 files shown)


ehfd (approved):
LGTM, as long as it was tested.

Issue #28, opened 2022-04-16
Mouse movement or speed is erratic in games (especially Wine) / Support relative mouse movement without uinput


xhejtman:
I have two issues with mouse pointer:

1. if pointer is not locked, mainly games cannot be controlled as mouse jumps randomly over screen. Could be related to resolution change? It does not happen e.g., with tuxracer. But it does happen in witcher 2.

2. if pointer is locked and remote and local resolution is significantly different, e.g., 4k vs. 1920x1080, pointer motion is very slow, acceleration should be adjusted.

ehfd:
+1
**Mouse speed is quite fast in Selkies by default, and it should definitely be slower.**

Normally this needs to be done from the application, not the system or container.
Some games or graphics application have these mouse speed scaling/mouse sensitivity options, but not all of them.
Use https://github.com/selkies-project/selkies-vdi/tree/master/images/uinput-device-plugin for now.
However, would be better mouse speed scaling/sensitivity was possible just with JavaScript and Xorg. It should be possible after how Selkies achieved without uinput.

ehfd:
Relevant code: https://github.com/selkies-project/selkies-gstreamer/blob/master/src/selkies_gstreamer/gstwebrtc_app.py and https://github.com/selkies-project/selkies-gstreamer/blob/master/src/selkies_gstreamer/webrtc_input.py

ehfd:
To be closed by #60.

ehfd:
I think that there should be a way for mouse behavior to work correctly even with the current configuration.

> if pointer is locked and remote and local resolution is significantly different, e.g., 4k vs. 1920x1080, pointer motion is very slow, acceleration should be adjusted.

I suspect that this is because the mouse cursor movements should be compensated for this resolution difference. This explains why pointer mode is slower or faster when the remote vs local resolution is different.

Communication between the screen size of the web interface and the screen size of the server to calculate the precise mouse movement intensity and sending to the X server with Xlib should be possible, and this may solve this issue without relying on uinput.

> if pointer is not locked, mainly games cannot be controlled as mouse jumps randomly over screen. Could be related to resolution change? It does not happen e.g., with tuxracer. But it does happen in witcher 2.

This is definitely related to relative mouse movement being sent as absolute mouse movements instead. I think going to pointer locked mode for 3D games such as The Witcher 2 is definitely adequate.

[...]

PR #95, opened 2023-09-20
Import of Selkies JS Interposer for WebRTC to Linux Joystick support

### LINKED ISSUE #55: Find a way to use Gamepad controllers without external container privileges
Currently, selkies-gstreamer does support gamepads only when the uinput device is provisioned. This is very cumbersome and risky in Kubernetes.

Instead of using that, it may be possible to directly translate them to Xorg operations (and later Wayland #46 using [virtual keyboard](https://wayland.app/protocols/virtual-keyboard-unstable-v1)).

Relevant files: [gst-web](https://github.com/selkies-project/selkies-gstreamer/tree/master/addons/gst-web/src) module and [webrtc_input.py](https://github.com/selkies-project/selkies-gstreamer/blob/master/src/selkies_gstreamer/webrtc_input.py).

### LINKED ISSUE #90: Add instructions to use uinput for bare metal/Docker instances, not just GKE or Kubernetes
The uinput interface is currently undocumented for anything other than uinput-device-plugin (for Kubernetes). Docs/adequate code should be added for normal OS or Docker interfaces.

danisla:
Fixes #90
Fixes #55

## New features
Import of Selkies JS Interposer for WebRTC to Linux Joystick support without uinput dependency
- LD_PRELOAD syscall interposer to emulate gamepads.
- Intercepts open() requests to /dev/input/jsX devices.
- Interposer connects to selkies unix domain socket.
- Interposer writes log to /tmp/selkies_js.log
- Selkies WebRTC app sends Javascript gamepad events to socket.
- Up to 4 simultaneous gamepads supported from WebRTC.
- Build as debian package and installed to /usr/local/lib/selkies-js-interposer/joystick_interposer.so
- Dev container updated to auto-load interposer.
- Update to github workflow to build and publish image and release asset.
- Add joystick interposer to example Dockerfile

Add turnserver to example docker image
- Helps with initial connectivity in environments like WSL with Docker.
- Can be overridden with env vars or cli args.

## Supported Gamepads
I've tested this with apps like retroarch and it seems to work well.

Currently only Xpad like gamepads are emulated for the button and axes mapping. 

I've tested with the following controllers:
- Xbox Series X/S Controller
- Stadia Controller
- 8Bitdo SN30 PRO

Others will surely work just fine, especially if they have similar button layouts to an Xpad.

Please let me know if there is a gamepad that doesn't work and we can try to accommodate.

## Testing
The `testing` branch has been built with the new joystick interposer pre-installed and added to the default LD_PRELOAD environment variable. This branch also contains the integrated turnserver to the example container, hence the new port 3478 to the docker run command.

You can test it with the example docker container like this:

Ubuntu 22.04
```
docker run --pull=always --name selkies -it --rm -p 8080:8080 -p 3478:3478 ghcr.io/selkies-project/selkies-gstreamer/gst-py-example:testing-ubuntu22.04
```

For Ubuntu 20.04:
```
docker run --pull=always --name selkies -it --rm -p 8080:8080 -p 3478:3478 ghcr.io/selkies-project/selkies-gstreamer/gst-py-example:testing-ubuntu20.04
```

## Usage
1. Connect to the web interface.
2. Connect a gamepad to your local computer.
3. Press any button on the gamepad to connect it to the web interface.
4. In the remote session, run a program like `jstest-gtk` to verify you see the `Selkies Controller`
5. Connect up to 4 total gamepads using the same method.
Selkies Gamepad Test

The example docker image has these environment variables set:
```bash
export LD_PRELOAD=/usr/local/lib/selkies-js-interposer/joystick_interposer.so
export SDL_JOYSTICK_DEVICE=/dev/input/js0
```
The entrypoint script also creates fake jsX files in /dev/input like below to help programs that try to list the /dev/input directory discover the joystick devices:
```
sudo mkdir -p /dev/input
sudo touch /dev/input/{js0,js1,js2,js3}
```

Some software programs use different methods to discover linux joysticks. The LD_PRELOAD method seems to work for most direct linux applications and the `SDL_JOYSTICK_DEVICE` env var helps with SDL applications.


ehfd:
Wow. This actually works?...

One aspect of testing that should be done is with Wine games and applications.

ehfd:
https://github.com/selkies-project/selkies-gstreamer/issues/28#issuecomment-1277196623

Could this approach be expanded to allow true relative mouse cursors?

danisla:
> Wow. This actually works?...
> 
> One aspect of testing that should be done is with Wine games and applications.

If someone can help test, that would be great.

danisla:
> [#28 (comment)](https://github.com/selkies-project/selkies-gstreamer/issues/28#issuecomment-1277196623)
> 
> Could this approach be expanded to allow true relative mouse cursors?

I think the approach would be similar, yes. My plan is to investigate doing this for mouse emulation later so we can completely remove the dependency on uinput and have full support in container environments.
However, Xorg has a funny way of loading core input devices like Mice and Keyboards so it might have to be a custom X11 module or a full uinput interposer. Either way, that's a separate task from the gamepad support in this PR.

ehfd:
Good to know that the approach is similar with mouse devices.

Unfortunately, I don't have a real gamepad device at the moment and I am having a hard time with time management.

@xhejtman perhaps interested?
Or any of the community.

src/selkies_gstreamer/__main__.py
```diff
@@ -383,9 +383,18 @@ def main():
     parser.add_argument('--uinput_mouse_socket',
                         default=os.environ.get('UINPUT_MOUSE_SOCKET', ''),
                         help='Path to the uinput mouse socket provided by the uinput-device-plugin, if not provided uinput is used directly.')
-    parser.add_argument('--uinput_js_socket',
-                        default=os.environ.get('UINPUT_JS_SOCKET', ''),
-                        help='Path to the uinput joystick socket provided by the uinput-device-plugin, if not provided uinput is used directly.')
+    parser.add_argument('--js0_socket',
```
ehfd: I think there is something better than to add four command line arguments. Would prefer that one command line argument that parses a maximum of four sockets.

src/selkies_gstreamer/webrtc_input.py
```diff
@@ -650,7 +677,8 @@ def on_message(self, msg):
             self.on_set_enable_audio(enabled)
         elif toks[0] == "_arg_resize":
             if len(toks) != 3:
-                logger.error("invalid _arg_resize commnad, expected 2 arguments <enabled>,<resolution>")
+                logger.error(
+                    "invalid _arg_resize commnad, expected 2 arguments <enabled>,<resolution>")
```
ehfd: Typo.

ehfd (commented):
Overall, looks okay but was just skimming. Made comments.

ehfd:
This commit is a great milestone which is a step forward from all known Linux remote desktop implementations working in containers.

PR #96, opened 2023-09-22
Video and audio in separate webrtcbins

### LINKED ISSUE #7: High latency when audio is enabled.

danisla:
Fixes #7 

- Reduces latency when streaming with audio enabled.
- AV sync is within reason.
- Client connects separate RTCPeerConnections for audio and video.
- Connections use the same config and STUN/TURN servers.
- Backend signal clients and pipelines are also duplicated.
- Removed option to disable audio from client and cli args.

## Testing

You can test it with the example docker container like this:

Ubuntu 22.04
```
docker run --pull=always --name selkies -it --rm -p 8080:8080 -p 3478:3478 ghcr.io/selkies-project/selkies-gstreamer/gst-py-example:testing-ubuntu22.04
```

For Ubuntu 20.04:
```
docker run --pull=always --name selkies -it --rm -p 8080:8080 -p 3478:3478 ghcr.io/selkies-project/selkies-gstreamer/gst-py-example:testing-ubuntu20.04
```

### DIFF
--- a/addons/gst-web/src/app.js
+++ b/addons/gst-web/src/app.js
@@ -76,7 +76,6 @@ var app = new Vue({
                 { text: '60 fps', value: 60 },
                 { text: '100 fps', value: 100 },
             ],
-            audioEnabled: false,
             audioBitRate: 32000,
             audioBitRateOptions: [
                 { text: '32 kb/s', value: 32000 },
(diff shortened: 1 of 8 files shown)


ehfd:
I am no longer making approvements in the meantime because I don't think I can test these in my limited time.

Let's just merge quickly and sort out the issues later.

PR #97, opened 2023-09-23
Fix relative mouse motion for most cases.

### LINKED ISSUE #28: Mouse movement or speed is erratic in games (especially Wine) / Support relative mouse movement without uinput

danisla:
Fixes #28

Changes how relative mouse movement is passed to X11 server using XTest extension.

pynput was not handling this correctly so this should fix it in most cases.

Tested this with gzdoom, which was previously broken and experienced the erratic mouse behavior in issue 28. Fix should be consistent for other cases as well.

### DIFF
--- a/src/selkies_gstreamer/webrtc_input.py
+++ b/src/selkies_gstreamer/webrtc_input.py
@@ -19,8 +19,9 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+import Xlib
 from Xlib import display
-from Xlib.ext import xfixes
+from Xlib.ext import xfixes, xtest
 import asyncio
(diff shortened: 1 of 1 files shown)


ehfd (approved):
I believe it should be better than before, even if we face something unexpected.

PR #101, opened 2023-10-02
exit if either of the two values doesn't exist or empty

PMohanJ:
A small fix at the if condition. As we need both values, turn_host and turn_port, we want the negation (not) to be applied to entire expression rather than just for the first variable followed by 'not' keyword.


ehfd (approved):
Looks good. Thank you.

PR #105, opened 2023-10-04
fix the split method

PMohanJ:
A small fix at the split method. If the variable [meta](https://github.com/selkies-project/selkies-gstreamer/blob/main/src/selkies_gstreamer/signalling_web.py#L356) is None then splitting the [message](https://github.com/selkies-project/selkies-gstreamer/blob/main/src/selkies_gstreamer/webrtc_signalling.py#L173) with `split(" ")` would return a list of length two with second variable as empty string `""`, thus raising the json decode error.

Screenshot 2023-10-04 at 3 52 41 PM

I've tested this change by creating a python build and now the error is gone.


ehfd (approved):
Good from me. No visible issues.

Issue #85, opened 2023-08-02
Reduce reliance on external libraries / certain upstream packages are unmaintained


ehfd:
https://github.com/selkies-project/selkies-gstreamer/commit/11006e393b9b7af81ab2b01d82a0092167da3e87

~~As specified here, the upstream python-uinput package is not maintained for 6-7 years.~~

~~Might need a change in the `uinput` interface to a project that continues to update, or maintain a selkies-project fork.~~

Moreover, why can't we completely replace `pynput.mouse.Controller()` and `pynput.keyboard.Controller()` with `Xlib.xtest`?

Also, we need to update all the dependencies for JavaScript.

ehfd:
We fork the repository in https://github.com/selkies-project/python-uinput

ehfd:
Merging with #90.

danisla:
> Might need a change in the `uinput` interface to a project that continues to update, or maintain a selkies-project fork.

What is the replacement project?

ehfd:
~~We may prefer to replace `python-uinput` with `python-evdev`, as we have gotten rid of `/dev/uinput` dependencies.~~

~~As far as I see, our reliance on `python-uinput` is limited to `uinput.BTN_LEFT`, `uinput.BTN_MIDDLE`, `uinput.BTN_RIGHT`, `uinput.REL_X`, `uinput.REL_Y`, `uinput.REL_WHEEL`, in `webrtc_input.py`. We aren't creating a new virtual input device with `/dev/uinput`, and our current usage should be covered with `python-evdev`.~~

Moreover, why can't we completely replace `pynput.mouse.Controller()` and `pynput.keyboard.Controller()`, and just use `Xlib.xtest` like we did in: https://github.com/selkies-project/selkies-gstreamer/blob/61f6b6c834337a02a9ad264fbd1be237418ce5f8/src/selkies_gstreamer/webrtc_input.py#L273-L277

ehfd:
https://github.com/selkies-project/selkies-gstreamer/pull/113

`python-xlib` is also unmaintained by the maintainer.

ehfd:
I am generally not too happy with dependencies: ~~`python-uinput` (dependency is effectively removed but still persists, plus it is unmaintained),~~ `pynput` (capabilities can be replaced by directly using `python-xlib`), `basicauth` (capabilities are trivial and replaceable with default packages), `gputil` (unmaintained and easily replaceable with another package such as `pynvml` and `gpustat`).

[...]

PR #113, opened 2023-11-23
Use version 0.31 of python-xlib to fix regression

cruizba:
Hello, congrats on the project! I've been testing Selkies and I'm impressed how well it works. VNC is slow for remote desktops, and this actually demonstrates that it's possible to have a low-latency remote desktop experience using WebRTC :)

As a Spanish user, I've been troubled with some keyboard layout errors, which significantly impacts my experience. Basically, if you press a non english character, the selkies-gstreamer server crash with an exception. This issue appearead multiple times as I can see:

- [Issue 84](https://github.com/selkies-project/selkies-gstreamer/issues/84)
- [Issue 9](https://github.com/selkies-project/selkies-gstreamer/issues/9)

Initially, I suspected that the problem might be due to guacamole keyboard library being outdated, as discussed in these threads:

- [Issue 9 Comment](https://github.com/selkies-project/selkies-gstreamer/issues/9#issuecomment-1705831611)
- [Issue 108](https://github.com/selkies-project/selkies-gstreamer/issues/108)
- [Issue 84 Comment](https://github.com/selkies-project/selkies-gstreamer/issues/84#issuecomment-1705831509)

However, even after updating the guacamole keyboard library, the issue persisted. Further research led me to discover a regression in the xlib python library, specifically since version 0.32:

- [Python-xlib Issue 241](https://github.com/python-xlib/python-xlib/issues/241)
- [Python-xlib Pull Request 242](https://github.com/python-xlib/python-xlib/pull/242)
- [Python-xlib Issue 259](https://github.com/python-xlib/python-xlib/issues/259)

If you execute a `pip list` command, the installed python-xlib version is 0.33. Fortunately, reverting the xlib version to 0.31 resolves the issue, and the server does not crash when a non english character is sent. What I did is basically pin the version 0.31 in the `setup.cfg` config, which could temporarily solve the issue until a new version of python-xlib is released. 

There is a pending pull request ([PR 242](https://github.com/python-xlib/python-xlib/pull/242)) that aims to address this problem in the xlib library, but it looks like the development is paused.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ install_requires =
     watchdog
     Pillow
     python-uinput @ git+https://github.com/selkies-project/python-uinput.git@0.11.3
-    python-xlib
+    python-xlib==0.31
 
 [options.packages.find]
 where = src


ehfd:
Hi!

I was pretty sure I knew your handle, and you're the one who made ubuntu-dind that I used with nvidia-dind!

This is indeed a tricky problem to deal with. I'll think of the implications of pegging this version before merging or adding to this.

cruizba:
Hahaha, it's such a small world! You created a project inspired by mine, and now I'm doing research inspired by your project 😂.

> Dealing with this is indeed a tricky problem. I'll consider the implications of pegging this version before merging or adding to it.

Yes, at first I thought it would be more difficult to do.

ehfd:
I remember that something would break if I downgraded to 0.31.
Could you confirm whether Selkies starts without major repercussions? @cruizba

cruizba:
I will close this PR. I think I was too optimistic with the downgrade...

It is a much more complex problem.

> Could you confirm whether Selkies starts without major repercussions? @cruizba
It starts correctly. The only difference I've noticed is that, by entering non english characters, the server does not crash using version 0.31

It starts correctly, but the keyboard still have issues:

---

**Case 1**
- `python-xlib=0.33`
- Host with keyboard layout `es` (Spanish).
- Container with keyboard layout `en-us` (Default)

In this particular case, the server crashes when I press "ñ" or other spanish characters.

---

**Case 2**
- `python-xlib=0.33`
- Host with keyboard layout `en-us` (US English).
- Container with keyboard layout `en-us` (Default)

No crashes

---

**Case 3**:
- `python-xlib=0.31`
- Host with keyboard layout `es` (Spanish).
- Container with keyboard layout `en-us` (Default)

No crashes. But the keyboard layout does not work properly. Some keys are translated, some not.

---

**Case 4**:
- `python-xlib=0.31`
- Host with keyboard layout `es` (Spanish).
- Container with keyboard layout `es` (Default)


Works fine, but also some characters are not translated correctly. For example, in Spanish layout, Ctrl + Shift + 0 resolves to the character "=", but instead, a ")" character appears.

---

**Some side notes**

This looks complex, and maybe other approach is needed, I am not sure.

I've observed that other projects like Kasm NoVNC translates the host keyboard layout, so if I have my host keyboard layout as `es` (Spanish), the keys are translated to the remote desktop, even if the remote keyboard is configured as `en-us`. Is there some kind of translation I am missing?

Would be good to research how other projects works (like neko, noVNC and Kasm noVNC), and somehow translate that to selkies :)

ehfd:
Yes, a comparison between other interfaces are needed. We currently use an old version of the Guacamole keyboard.

ehfd:
The last time there was some keyboard issue, it was related to pynput. Worth checking into the backend aspects if the Guacamole Keyboard behavior is inconsistent with guacd.

maxpain:
Any updates?

ehfd:
Currently discussed in #85

PR #114, opened 2023-11-23
Fix devcontainer for ubuntu jammy

cruizba:
VSCode dev containers was broken for ubuntu jammy images because of a missing `apt-get update` before installing `xcvt`.

This fixed the issue, and I was able to start doing things after the fix :)


ehfd:
Merged, thanks a million!

PR #134, opened 2024-01-11
Update docker commands to utilize coturn properly

PMohanJ:
As the coturn is included in the desktop itself, when deploying the container in a VM rather than local env, the `TURN_HOST` variable is supposed to set to the host address of the machine, else the default value `localhost` is applied which in these scenarios fails the webrtc connection.

Also when setting up the coturn server with the provided docker command it requires the realm flag/argument to set for the TURN capability to work https://github.com/coturn/coturn/blob/master/docker/coturn/turnserver.conf#L352.  

This can be verified by doing an [ICETEST](https://icetest.info/). 

```bash 
docker run -d -p 3478:3478 -p 3478:3478/udp -p 49160-49200:49160-49200/udp coturn/coturn -n --min-port=49160 --max-port=49200 --lt-cred-mech --user=yourusername:yourpassword
```

```bash
docker run -d -p 3478:3478 -p 3478:3478/udp -p 49160-49200:49160-49200/udp coturn/coturn -n --min-port=49160 --max-port=49200 --lt-cred-mech --user=yourusername:yourpassword --realm=example.com
```

Run the both containers and test them for ICE candidates at the given site and check for the `relay` candidates. The coturn without the realm flag doesn't return `relay` ICE candidate.

### DIFF
--- a/Dockerfile.example
+++ b/Dockerfile.example
@@ -210,7 +210,7 @@ sudo /usr/bin/pulseaudio --daemonize --system --verbose --log-target=file:/tmp/p
 sudo /start-turnserver.sh &\n\
 export WEBRTC_ENCODER=\${WEBRTC_ENCODER:-x264enc}\n\
 export WEBRTC_ENABLE_RESIZE=\${WEBRTC_ENABLE_RESIZE:-true}\n\
-export TURN_HOST=\${TURN_HOST:-localhost}\n\
+export TURN_HOST=\${TURN_HOST:-$(curl checkip.amazonaws.com)}\n\
 export TURN_PORT=\${TURN_PORT:-3478}\n\
 export TURN_USERNAME=\${TURN_USERNAME:-selkies}\n\
 export TURN_PASSWORD=\${TURN_PASSWORD:-selkies}\n\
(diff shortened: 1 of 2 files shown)


ehfd:
I acknowledge that both may be an issue. But, for the `TURN_HOST`, it would be a better idea to identify the IP from within the example container. There's `checkip.amazonaws.com` and alternatives that are typically used in this case.

PMohanJ:
I tried with `checkip.amazonaws.com` it's returning the Public IP of the VM, but in case of VPN enabled VMs it's not working because the client can only access that VM with the private IP but not Public IP (correct me if I'm wrong, latter is my scenario here and it's not working for me).

ehfd:
In that case, it would be possible to add that detail as a note and default to using automatic IP resolution.

PMohanJ:
Sounds good.

ehfd:
@PMohanJ Care to try making that change? I'll make the final edits tomorrow.

PMohanJ:
Ok. I'll update that in a few minutes.

PMohanJ:
@ehfd Any review on this?

ehfd:
Give me a bit of time.
Thank you for your patience.

ehfd:
Thank you.

Issue #139, opened 2024-04-03
Implement WebRTC statistics logging in host


ehfd:
https://gstreamer.freedesktop.org/documentation/webrtc/index.html?gi-language=python#webrtcbin::get-stats

The above function allows the host to obtain WebRTC Stats API outputs.

Despite this being an enhancement, I'm tagging this as urgent because analysis is impossible without this.

ehfd:
Should be integrated with https://github.com/selkies-project/selkies-gstreamer/blob/main/src/selkies_gstreamer/metrics.py

The metrics module should support toggling between file logging (preferably a CSV) and Prometheus port logging.

ehfd:
Completed in #141

PR #141, opened 2024-04-06
Logging webrtc statistics on host received over datachannel

PMohanJ:
This is a DRAFT: Do not merge it yet.

The reports are being sent over datachannel and being dumped to a CSV file at `/tmp/stats.csv` location.
Due to duel webrtc connection the audio stats aren't included yet as the file size keeps increasing rapidly even with just video webrtc api stats.

#139

### DIFF
--- a/addons/gst-web/src/app.js
+++ b/addons/gst-web/src/app.js
@@ -474,8 +474,11 @@ function onBothStreamConnected() {
 
                 statsStart = now;
 
-                var jsonString = JSON.stringify(stats.allReports);
-                webrtc.sendDataChannelMessage("_stats," + jsonString)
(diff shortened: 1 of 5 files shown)


ehfd:
File size is probably expected.

ehfd:
Closes #139

PMohanJ:
So for audio stats, should they be in a separate csv file?

ehfd:
That would be desirable.

PMohanJ:
The changes include logging of webrtc statistics of both audio and video reports. Also the logging of stats is per session based, meaning that for every new session new stats files are generated. This is done to maintain the structure of csv file as different browsers provide statistics in different order.

ehfd (approved):
I am happy with the changes.

PR #156, opened 2024-05-23
Fix Flash of Unstyled Content (FOUC) issue

vishalkadam47:
- The addition of the v-cloak directive and the accompanying CSS rule will help to prevent the application from displaying partially loaded or unstyled content while it's loading.

- The Flash of Unstyled Content (FOUC) is likely caused by the way the Vue.js application is initialized and rendered. The root cause is that the HTML content is rendered before the Vue.js application is mounted and the styles are applied.

### DIFF
--- a/addons/gst-web/src/index.html
+++ b/addons/gst-web/src/index.html
@@ -16,6 +16,10 @@
       font-family: Roboto, Arial, sans;
     }
 
+    body {
+      background-color: #000000;
+    }
+
     .scrolly textarea {
       min-height: 300px;
       white-space: pre;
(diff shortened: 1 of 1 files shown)


addons/gst-web/src/index.html
```diff
@@ -123,12 +123,13 @@
       word-wrap: normal;
       direction: ltr;
     }
+    body { background-color: #000000; }
```
PMohanJ: Format the code as following:
```
body {
    background-color: #000000;
}
```

addons/gst-web/src/index.html
```diff
@@ -385,6 +386,9 @@
       </div>
     </v-app>
   </div>
+  <style>
+    [v-cloak] { display: none; }
```
PMohanJ: Include this style rule in the style tag of head section. Format it as following:
```
[v-cloak] {
      display: none; 
}

PMohanJ (changes requested):
Kindly make the changes as described.

PMohanJ (approved):
Looks Good To Me.

ehfd (approved):

PR #162, opened 2024-07-26
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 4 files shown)


ehfd:
Thanks for your recommendation, and I'm putting information about this on the `development.md` style guide as a recommended procedure for contributions.

However, it seems that codespell is not 100% accurate at all times, requiring manual approval, and since it's going to have read permissions only, I would prefer to leave it for individual developers to run in their pull requests instead of the CI.

ehfd:
I've fixed all the typos thanks to your PR and added adequate instructions about using codespell for developers. Because I'd like to constrain CI/CD pipelines to those that are automatic, I'll postpone the PR which adds the CI for now.

PR #189, opened 2025-07-22
bugfix - typo in initial logic this disables streaming mode by default

thelamer:
This was a typo, the default should be false and this keeps it in sync with the user setting.


thelamer:
Merging without approval due to simplicity of change.

PR #204, opened 2025-09-23
add a RO websocket with no input handling

### LINKED ISSUE #203: Add a RO  websocket, enforce it on the backend instead of the frontend
**Is your enhancement request related to a problem? Please describe.**

In more secure environments, we need to guarantee the read-only mode is truly read-only - meaning only the person who "owns" the desktop can send keystrokes.

The current read-only endpoint is enforced via the JS frontend. Keystrokes will still be accepted by the open websocket.

**Describe the solution you'd like**
I would like to have a websocket endpoint that is guaranteed to be RO on the backend and is not just guarded by the frontend.

It can then be guarded by my ingress/reverse proxy with its own authentication, allowing me to plug in the correct authorization for my use case/deployment/organization etc.

Since in most modern envs such things are handled outside the application, I think it'd be great to keep it lean and allow the existing infrastructure to handle authentication and routing the traffic here - all that'd need is for Selkies to listen on a separate port where the input is ignored.

By simply adding another DataStreamingServer with a None input_handler, I managed to get this working.

I'll submit a PR - do let me know if there are better ways to solve this however, I'm happy to contribute here, as we use Selkies with similar use cases quite heavily.

This issue loosely relates to https://github.com/selkies-project/selkies/issues/39 , but doesn't implement it fully.

 - [ x ] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
 - [ x ] I confirm that I have read other open and closed issues and that duplicates do not exist.
 - [ x ] I confirm that I have described the solution as well as the problem in detail, and if possible, explained how to solve the problem.
 - [ x ] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [ x ] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

danielpodwysocki:

Closes #203 

The goal here is to provide a mechanism to ensure a view-only connection, per issue #203 .

Currently, this is handled on the frontend, meaning as long as the websocket is open, it is possible to send keystrokes through.

Adding another endpoint allows for clean separation and smooth integration with reverse proxies/middlewares/auth setups out there.

This helps a use case where a user has got an established setup around that or wants to integrate with their platform - so custom auth middlewares can smoothly handle routing users between RW/RO endpoints.

This adds a RO DataStreamingServer on a separate port. It sets None as its input handler, meaning all inputs are skipped. 

That was the only way I could figure out that prevents sending keystrokes and is not susceptible to misuse/malicious actors. 

**Additional context**

 - [ x ] I confirm that this pull request is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please file the pull request there.
 - [ x ] I confirm that this pull request has been tested thoroughly and to the best of my knowledge that additional unintended problems do not arise.
 - [ x ] I confirm that the style of the changed code conforms to the overall style of the project.
 - [ x ] I confirm that I have read other open and closed pull requests and that duplicates do not exist.
 - [ x ] I confirm that I have justified the need for this pull request and that the changes reflect the fix for the specified problem.
 - [ x ] I confirm that no portion of this pull request contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [ x ] I confirm that the authors of this pull request does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

### DIFF
--- a/src/selkies/selkies.py
+++ b/src/selkies/selkies.py
@@ -3339,7 +3339,7 @@ async def _send_stats_periodically_ws(websocket, shared_data, interval_seconds=5
     except Exception as e:
         data_logger.error(f"Stats sender (WS) error: {e}", exc_info=True)
 
-async def on_resize_handler(res_str, current_app_instance, data_server_instance=None, display_id='primary'):
+async def on_resize_handler(res_str, current_app_instance, data_server_instance=None, display_id='primary', ro_data_server_instance=None):
     """
     Handles client resize request. Updates the state for a specific display and triggers a full reconfiguration.
     """
@@ -3456,6 +3456,12 @@ async def main():
         type=int,
         help="The port for the data websocket server. Overrides the CUSTOM_WS_PORT environment variable.",
     )
+    parser.add_argument(
+        "--view_only_port",
+        default=os.environ.get("CUSTOM_RO_WS_PORT", "8083"),
+        type=int,
+        help="The port for the view-only data websocket server. Overrides the CUSTOM_RO_WS_PORT environment variable.",
+    )
     parser.add_argument("--debug", action="store_true", help="Enable debug logging")
     args, unknown = parser.parse_known_args()
     global TARGET_FRAMERATE, TARGET_VIDEO_BITRATE_KBPS
@@ -3504,6 +3510,19 @@ async def main():
         audio_device_name=args.audio_device_name,
         cli_args=args,
     )
+    ro_data_server = DataStreamingServer(
+        port=args.view_only_port,
+        app=app,
+        uinput_mouse_socket=UINPUT_MOUSE_SOCKET,
+        js_socket_path=JS_SOCKET_PATH,
+        enable_clipboard=ENABLE_CLIPBOARD,
+        enable_cursors=ENABLE_CURSORS,
+        cursor_size=CURSOR_SIZE,
+        cursor_scale=1.0,
+        cursor_debug=DEBUG_CURSORS,
+        audio_device_name=args.audio_device_name,
+        cli_args=args,
+    )
     app.data_streaming_server = data_server
 
     input_handler = InputHandler(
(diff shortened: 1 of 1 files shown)


src/selkies/selkies.py
```diff
@@ -3504,6 +3510,19 @@ async def main():
         audio_device_name=args.audio_device_name,
         cli_args=args,
     )
+    ro_data_server = DataStreamingServer(
+        port=args.view_only_port,
+        app=app,
+        uinput_mouse_socket=UINPUT_MOUSE_SOCKET,
+        js_socket_path=JS_SOCKET_PATH,
+        enable_clipboard=ENABLE_CLIPBOARD,
```
danielpodwysocki: This should probably be false, to ensure we can't "spy on" the clipboard.

danielpodwysocki: Updated.

thelamer:
I completely missed this PR and I apologize. 
Just in general this can be achieved with the token authentication method now: 
https://github.com/linuxserver/docker-baseimage-selkies/?tab=readme-ov-file#control-plane-api-for-token-management

Is this good enough or is auth free still needed.

hitmanmotors2-coder:
I'm new to this so I apologize if this is the wrong place, but part of the issue is also that the shared link is root path with "#shared" added to the end.  Since anchors aren't sent from client to server, there is no easy way for a reverse proxy to force certain clients to #shared and allow others to the read/write path.  Ideally you could setup a caddy or other reverse proxy with 2 urls and use the path to direct one client to the write path and the other to the read only path (or apply auth to one side).  

Another small problem, if auth is turned on at the selkies container, it pops up for both the write and the read only page.  So you cant apply auth to only the write portion and allow the read only to be unauthenticated and as mentioned above.   Ant the reverse proxy cant do this due to the use of an anchor rather then separate paths.

Both would be helpful I think in this enhancement.  This would make a great wallboard display, for example, if only one person could login and display metrics (such as call metrics) and the read only link could be displayed on monitors within an office, if there was a way to get a reverse proxy to split the write and read only paths.

ehfd:
I agree that an RO WebSocket path is nice, but I am also against a separate port for view-only. I prefer the same port to handle this.

PR #207, opened 2025-09-29
feat(dashboard): add new selkies-dashboard-wish wrapped everything in a single top bar

vishalkadam47:
- Add network bandwidth and latency monitoring to system stats
- Implement touch gamepad and trackpad mode support
- Enhance clipboard with image upload functionality
- Add support for second screen placement
- Update UI title and logo visibility based on server settings


thelamer (dismissed):
The readme needs to be cleaned up
It looks like you have leftover images from the previous dash

It is not a hard requirement but the updated code selectively renders components based on if the settings are clamped or locked. 

Looks slick though and runs well.

vishalkadam47:
Thanks for the feedback @thelamer. Glad to hear it’s running well

PR #224, opened 2026-03-22
docs: add Firefox clipboard workaround to FAQ

### LINKED ISSUE #202: Copy + paste not working in Firefox
Direct copy + paste does not work in Firefox (v143.0) for me unless I use the Server Clipboard field from Selkies Sidebar. In Chrome it does work after allowing Clipboard access (note that Firefox never asks for Clipboard access).

After looking into this, the clipboard does work after setting `dom.events.testing.asyncClipboard = true`  in `about:config`.

Maybe this fix/workaround could be added to the troubleshooting FAQ? https://selkies-project.github.io/selkies/faq/#the-clipboard-does-not-work

Using Selkies through a linuxserver baseimage: https://github.com/linuxserver/docker-baseimage-selkies/blob/master/Dockerfile

--edit--
Sorry, this is probably not a bug...

aliefe04:
Fixes #202

## Summary

Added Firefox-specific workaround to the clipboard FAQ section.

## Problem

In Firefox, the clipboard does not work even with HTTPS because Firefox does not prompt for clipboard access like Chrome does.

## Solution

Document the workaround: enable `dom.events.testing.asyncClipboard` in `about:config`.

## Testing

The workaround was verified by the issue reporter in #202.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -55,6 +55,8 @@ However, it might be that the parameters for the WebRTC interface, video encoder
 
 This is very likely a web browser constraint that is applied because you are using HTTP for an address to the web interface that is not localhost. The clipboard only works when you use HTTPS (with a valid or self-signed certificate), or when accessing localhost (some browsers do not support this as well). You could use port forwarding to access through localhost or obtain an HTTPS certificate.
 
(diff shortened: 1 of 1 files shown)


ehfd:
I'm going to merge it but the docs will be replaced.

ehfd (approved):

PR #226, opened 2026-03-24
fix: replace bare except with except Exception in legacy modules

harshadkhetpal:
## Summary

Replace bare `except:` clauses with `except Exception:` in legacy gamepad and WebRTC input modules.

Bare `except:` catches all exceptions including `SystemExit`, `KeyboardInterrupt`, and `GeneratorExit`, which is almost never intended. Using `except Exception:` follows PEP 8 best practices.

**Changes:**
```python
# Before (gamepad.py — 2 occurrences)
except:
    pass

# After
except Exception:
    pass

# Before (webrtc_input.py — 4 occurrences)
except:
    logger.error(...)

# After
except Exception:
    logger.error(...)
```

## Files Changed
- `src/selkies/legacy/gamepad.py` — 2 bare except → except Exception
- `src/selkies/legacy/webrtc_input.py` — 4 bare except → except Exception

## Testing
No behavior change for normal exceptions — style/correctness fix only.

### DIFF
--- a/src/selkies/legacy/gamepad.py
+++ b/src/selkies/legacy/gamepad.py
@@ -328,7 +328,7 @@ class SelkiesGamepad:
             self.server.close()
             try:
                 os.unlink(self.socket_path)
-            except:
+            except Exception:
                 pass
         
         logger.info("Stopped gamepad socket server for %s" % self.socket_path)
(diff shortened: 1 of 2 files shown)


ehfd:
Sorry, we just decided an hour ago that we would remove the legacy GStreamer code...

PR #235, opened 2026-04-19
fix(input): Resolve unicode dropping, stabilize clipboard, and fix undo behavior.

athene20:
**Summary**
- Implement an asynchronous `unicode_buffer` to debounce rapid keystrokes during IME composition.
- Prevent race conditions under high CPU load by batching clipboard injections instead of executing them per keystroke.
- Intercept Backspace (keysym 65288) during composition to pop characters from the buffer rather than triggering a premature flush.
- Ignore `keyup` events during Unicode input to maintain composition context.
- Increase the post-paste delay in `_inject_unicode_via_clipboard` from 0.02s to 0.05s to ensure the target application reads the clipboard before restoration.
- Improve Undo (Ctrl+Z) behavior by preventing the target application's undo history from being polluted with intermediate IME backspaces.

https://github.com/linuxserver/docker-baseimage-selkies/issues/150
@thelamer 

### Before this PR
- **Per-Keystroke Injection:** Every single Unicode keydown triggered a heavy clipboard injection cycle (`backup -> write -> Shift+Insert -> restore`).
- **Composition Breakage:** The client simulates IME composition by sending sequences like `char -> backspace -> combined char`. The previous logic treated this `backspace` (keysym 65288) as a control key, immediately flushing the buffer and breaking the composition.
- **Polluted Undo History (Ctrl+Z):** Because every intermediate step and backspace during IME composition (e.g., `ㅇ` -> `Backspace` -> `아` -> `Backspace` -> `안`) was individually injected into the OS, the target application recorded every single step. Users had to press `Ctrl+Z` multiple times just to undo one composed character.
- **Unwanted Keyup Flushing:** `keyup` events generated during IME composition caused premature buffer flushes.
- **Premature Clipboard Restoration:** The post-paste wait time in `_inject_unicode_via_clipboard` was set to `0.02s`. Under high CPU load, this was too short, causing the clipboard to be restored to its original state *before* the target application could actually read the injected text.

### After this PR (The Solution)
- **Debouncing with `unicode_buffer`:** Introduced an asynchronous buffer to debounce rapid Unicode keystrokes. The buffer waits for a short timeout (`0.15s`) before flushing, executing the heavy clipboard injection only once for a complete string (e.g., a full word) rather than per character.
- **Backspace Interception:** If a Backspace (`0xFF08`) is received while the `unicode_buffer` has data, it pops the last character from the buffer instead of flushing it to the screen. This perfectly syncs with the client's IME composition behavior.
- **Restored Natural Undo (Ctrl+Z) Behavior:** Since the text is now injected as a complete, debounced batch after the user finishes composing a word, the target application registers it as a single, atomic action. This restores the expected, single-step Undo functionality.
- **Ignored Unicode Keyups:** `keyup` events for Unicode fallbacks are now safely ignored, ensuring the buffer only flushes on actual context-breaking keys (Enter, Space, etc.) or timeout.
- **Stabilized Injection Timing:** Increased the post-paste delay from `0.02s` to `0.05s` to guarantee the target application has enough time to process the `Shift+Insert` action before the clipboard is restored.

### Key Changes
- Refactored `_keyboard_worker` in `input_handler.py`.
- Added `flush_buffer()` logic with a `0.15s` `asyncio.wait_for` timeout.
- Added logic to safely intercept keysym `65288` (Backspace) for local buffer manipulation.
- Increased `asyncio.sleep(0.02)` to `asyncio.sleep(0.05)` after the `Shift+Insert` sequence in `_inject_unicode_via_clipboard`.

- Works as intended.
- No changes to dependencies.
- The input isn't immediate, so it seems like it's lagging. (We may need to consider adjusting 0.15 seconds to 0.1 seconds.)

I think the best approach is to pass the keys directly without using the clipboard, but I'm not sure if that's possible.

**Additional context**
- Before

https://github.com/user-attachments/assets/dd1474ad-ab56-47ba-a741-e6040a3807c4

- After

https://github.com/user-attachments/assets/525b215b-86ff-4249-bbe6-af43ebc1b956

### DIFF
--- a/src/selkies/input_handler.py
+++ b/src/selkies/input_handler.py
@@ -1419,7 +1419,7 @@ class WebRTCInput:
                 await self.send_x11_keypress(KEY_INSERT, down=True)
                 await self.send_x11_keypress(KEY_INSERT, down=False)
                 await self.send_x11_keypress(KEY_SHIFT_L, down=False)
-                await asyncio.sleep(0.02)
+                await asyncio.sleep(0.05)
 
                 if old_data is not None:
                     await self.write_clipboard(old_data, mime_type=old_mime or "text/plain")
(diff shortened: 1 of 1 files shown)


thelamer (approved):
Yeah this is a much better approach, passed my testing and is unobtrusive to the rest of the input stack. 

Nice work!

thelamer:
We ingest from selkies head at the end of every week so this will be in this upcoming weekends waterfall.

ehfd:
Thank you for the contribution! Anything further are also appreciated!

PR #250, opened 2026-06-10
Fix WebSocket reconnect cleanup and clipboard subprocess timeouts

rmounce:
## Summary

Fix cleanup issues exposed by repeated WebSocket reconnects in websockets mode.

This PR addresses two related problems:

- Per-connection stats monitor tasks were not cancelled on WebSocket disconnect because cleanup looked for local variable names that were never assigned.
- Clipboard subprocesses (`xclip` / `wl-*`) were not killed and reaped after timeout, and outbound clipboard polling continued even when no WebSocket clients were connected.

## Details

Each WebSocket connection starts system, GPU, stats sender, and network monitor tasks. The task handles were stored on `self`, but the connection cleanup checked for underscored names in `locals()`. Those names
did not exist, so reconnect churn could leave monitor loops running after disconnect. On GPU hosts this is especially visible because the GPU monitor calls `GPUtil.getGPUs()`, which shells out to `nvidia-smi`.

Clipboard reads and writes already used timeouts, but a timeout only cancelled the Python wait; it did not terminate and reap the subprocess. This PR wraps clipboard subprocess waits so timed-out commands are
killed and reaped before the timeout is propagated.

In websocket mode, the outbound clipboard monitor now also skips host clipboard polling while there are no connected WebSocket clients. The last-sent marker is reset while idle so the next client receives the
current clipboard state after reconnect.

## Testing

- `python3 -m py_compile src/selkies/selkies.py src/selkies/input_handler.py`
- `git diff --check upstream/main..HEAD`
- Reconnect stress testing against a production-like container showed stable task/subprocess counts after applying these fixes.

### DIFF
--- a/src/selkies/input_handler.py
+++ b/src/selkies/input_handler.py
@@ -1715,6 +1715,45 @@ class WebRTCInput:
             logger_webrtc_input.warning("Failed to access clipboard file %s: %s", file_path, e)
             return None, None
 
+    async def _kill_and_reap_process(self, proc, description):
+        logger_webrtc_input.warning(
+            "Timed out waiting for clipboard command '%s' pid=%s; killing it.",
+            description,
+            getattr(proc, "pid", "unknown"),
+        )
+        try:
+            proc.kill()
+        except ProcessLookupError:
+            pass
+        try:
+            await asyncio.wait_for(proc.wait(), timeout=1.0)
+        except (asyncio.TimeoutError, ProcessLookupError):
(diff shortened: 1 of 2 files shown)


ehfd:
Directly applied to https://github.com/selkies-project/selkies/commit/043ce6bfc840bffb8bfc46391724a177ec6f5f4e and https://github.com/selkies-project/selkies/commit/47d2c13463f909d2f353684a46a20c0b504e19de with you as the author. Thank you!

PR #251, opened 2026-06-10
Send current cursor to websocket clients on connect

rmounce:

N/A

WebSocket clients can initially show a blank cursor until the remote desktop emits a cursor-change event. The server already caches and replays the last cursor payload, but that cache can be empty or stale when a
new client connects.

This pull request refreshes the cursor cache from the current XFixes cursor image before sending the initial cursor payload to a new websocket client. It also sends the current cursor immediately after granting
mouse/keyboard access, so control handoff receives an accurate cursor after input attaches.

- Added `get_current_cursor_data()` to the input handler to snapshot the current X11 cursor through the existing XFixes and `cursor_to_msg()` path.
- Added websocket server helpers to refresh the cursor cache and send the current cursor to a specific client.
- Replaced initial “last cursor” replay with a fresh current-cursor send on websocket connection.
- After `MK_ACCESS,1`, sends the current cursor to the newly active input client.

Tested with:

- `python3 -m py_compile src/selkies/selkies.py src/selkies/input_handler.py`
- `git diff --check upstream/main..HEAD`
- Production-version branch testing confirmed the cursor appears correctly on initial connection/control handoff.

A frontend fallback was considered, where the browser default cursor is shown until a remote cursor update arrives. That is less accurate and may be undesirable for gaming or applications that intentionally hide
the cursor. Refreshing and sending the actual backend cursor is more accurate and avoids showing a browser default cursor over applications that have hidden their cursor.

**Additional context**

This change uses the existing cursor serialization path and does not introduce a new cursor format or client protocol message.

### DIFF
--- a/src/selkies/input_handler.py
+++ b/src/selkies/input_handler.py
@@ -1989,6 +1989,25 @@ class WebRTCInput:
         logger_webrtc_input.info("stopping cursor monitor")
         self.cursors_running = False
 
+    def get_current_cursor_data(self):
+        if self.is_wayland:
+            return None
+        if not self.enable_cursors or not self.xdisplay:
+            return None
+        try:
+            if not self.xdisplay.has_extension("XFIXES"):
+                if self.xdisplay.query_extension("XFIXES") is None:
+                    logger_webrtc_input.error(
+                        "XFIXES extension not supported, cannot fetch current cursor"
+                    )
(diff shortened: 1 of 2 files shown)


ehfd (approved):

ehfd:
Thank you for the contribution!

PR #245, opened 2026-05-07
feat(dashboard-wish): enhance sharing functionality and UI customization

vishalkadam47:
- Refactored Sharing component to compute renderable settings directly from server settings.
- Removed deprecated ShortcutsMenu component.
- Added video bitrate monitoring to SystemMonitoring component.
- Introduced PlayerGamepadButton for touch gamepad functionality in player modes.
- Updated TopMenu to conditionally render UI elements based on server settings.
- Improved Vite configuration for better build performance and environment handling.
- Added utility functions for computing renderable settings and route prefix.

PR #169, opened 2024-12-19
Added commands and variables to run something when client connect/disconnect

skipperro:
- Added support for SELKIES_START_AFTER_CONNECT and SELKIES_START_AFTER_DISCONNECT variables
- Added support for --start-after-connect and --start-after-disconnect start arguments
- Removed duplicated base64 import in signalling_web.py

**Additional context**

 - [x] I confirm that this pull request is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please file the pull request there.
 - [x] I confirm that this pull request has been tested thoroughly and to the best of my knowledge that additional unintended problems do not arise.
 - [x] I confirm that the style of the changed code conforms to the overall style of the project.
 - [x] I confirm that I have read other open and closed pull requests and that duplicates do not exist.
 - [x] I confirm that I have justified the need for this pull request and that the changes reflect the fix for the specified problem.
 - [x] I confirm that no portion of this pull request contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [x] I confirm that the authors of this pull request does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

Issue #227, opened 2026-03-27
[META] Road to v2.0.0 Release


ehfd:
Will be consistently updated. Entries without assignees are all open to pull requests. Vibe coding is welcome as long as functional validation is correct.

Checkmarked entries are currently in Pull Requests. Will be unchecked if it gets closed without merging, and will be struck out and moved to the bottom if merged.

## Done

- [x] ~~Expose CBR for WebSockets and CRB as an option for WebRTC mode. - Assignee: @PMohanJ~~
- [x] ~~https://github.com/selkies-project/selkies/issues/74 - Assignee: @PMohanJ~~
- [x] ~~https://github.com/selkies-project/selkies/issues/172 - Assignee: @ehfd~~
- [x] ~~Use one port for Data and WebSocket Streams, and embed an internal asyncio web server instead of NGINX (NGINX can be used to make things faster, optionally, but Selkies is self-contained otherwise). - Assignee: @PMohanJ~~
- [x] ~~Integrate the web components inside the Python wheel, not as a separate directory that has to be installed. - Assignee: @ehfd or @PMohanJ~~
- [x] ~~https://github.com/selkies-project/selkies/issues/147, expose https://github.com/linuxserver/pcmflux/pull/4 to Selkies - Assignee: @PMohanJ~~
- [x] ~~Bring NVENC with X11 back to working state in containers with multiple GPUs; Contact me for details, although it's likely I'll do it myself. - Assignee: @ehfd.~~
- [x] ~~Fix NVENC encoder dropping a lot of frames with Windows Chromium `D3D11VideoDecoder` - Consult settings in `https://github.com/selkies-project/selkies/blob/7dd6a93065332e5b3fe48fe5e5eb911a91f82746/src/selkies/legacy/gstwebrtc_app.py#L257-L938`. The decoder initially starts from `baseline` profile and changes to `high` (open Details in https://github.com/selkies-project/selkies/issues/227#issuecomment-4143090247).~~
- [x] ~~Integrate Python-Xlib code into Selkies and remove that repository (removes the ugly workaround in `pyproject.toml`). Abide by the LGPL license so that such code is only connected with Selkies using `import` statements, and leave the license file in the subdirectory.(https://github.com/selkies-project/selkies/issues/85) - Assignee: @ehfd~~
[...]

PR #254, opened 2026-06-13
fix(core): Comprehensive fixes and performance optimizations

### LINKED ISSUE #85: Reduce reliance on external libraries / certain upstream packages are unmaintained

### LINKED ISSUE #145: Keyboard keystrokes continue to be held when packets are lost by congestion or unstable connection / Improve the last bit of keyboard latency overall while accommodating unstable connections
Keys continue to be held when packets are lost by congestion or unstable connection. This is because key presses are sent by signals when they are pressed and when they are down.

https://github.com/Xpra-org/xpra/blob/a5a13a993139d9d48a60734e7bab3ec9d53c919a/xpra/server/mixins/input.py#L257

The above is a good solution to this problem, to send `key-repeat` packets continuously when a key-press is made and when such packets stop, un-press from the host.

### LINKED ISSUE #153: Improve Opus: reduce NetEQ jitterbuffer & eliminate discontinuous audio artifacts from jitter with Opus + RED
## NetEQ jitterbuffer

Video frames can be displayed as fast as they can be received as a result of `playout-delay`, `jitterBufferTarget`, and other WebRTC settings.

However, the minimum latency for Opus audio is constrained by NetEQ. This behavior is not well-researched, but when understood, this may decrease the remaining bits of latency the audio stack has.

https://chromium.googlesource.com/external/webrtc/+/master/modules/audio_coding/neteq/g3doc/index.md
https://bloggeek.me/webrtc-discard-media-packets/

Moreover, while I have specified 2.5 (3) ms as the frame size and ptime of the Opus packets, this might not have an actual effect. This behavior should be clearly verified. Edit all sections with `OPUS_FRAME:` to modify this.

## Opus + RED

**An issue required to be solved from upstream, tracked in:** https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/3550

Described below:

There is a lack of redundancy available with the audio stack, especially with Opus.
This causes concealment events which lead to disconnections in audio in all but the most reliable wired ethernet connections.

The `inband-fec` parameter of Opus doesn't work with `restricted-lowdelay` (CELT-only) configurations that most people would want when streaming any other type of media other than live voice (SILK).

An alternative (enabled by default with Chrome 96, other links specifying flags are before Chrome 96) is to use RED with Opus.

https://groups.google.com/g/discuss-webrtc/c/5761etCrSuA/m/hEJto5xwBgAJ

https://webrtc.github.io/samples/src/content/peerconnection/audio/

https://webrtchacks.com/red-improving-audio-quality-with-redundancy/

https://www.meetecho.com/blog/opus-red/

https://webrtchacks.com/implementing-redundant-audio-on-an-sfu/

This is what many providers (including Zoom) seem to use, and there are reports (above) that even in jittery conditions, RED with distance=2 improves concealment from 60% to 18% and 20% to 0.7%.

The current `GST_WEBRTC_FEC_TYPE_ULP_RED` is meant for video, and along with the string `red/48000` instead of `red/48000/2` that Opus requires, doesn't work overall. When manually munging the `red/48000/2` string into the SDP and using [`rtpredenc`](https://gstreamer.freedesktop.org/documentation/rtp/rtpredenc.html) with `request-fec-encoder`, RED packets are sent correctly, but those packets (2 times larger than the Opus bitrate with distance=2) are instead played by Chrome's WebRTC stack, not Opus, contrary to https://webrtc.github.io/samples/src/content/peerconnection/audio/.

I propose that this capability be called `GST_WEBRTC_FEC_TYPE_OPUS_RED`.

### LINKED ISSUE #157: [META] Optimize the WebRTC stack to the maximum
Linked with #153, #152, #34, #30

**Also read: https://github.com/m1k1o/neko/issues/371**

In the v1.6.0 release, there is much higher confidence in our performance optimizations in the WebRTC stack.
We have achieved a way to eliminate jitterbuffer latency from the WebRTC decoder using `playout-delay` and `jitterBufferTarget`, along with many other measures to stabilize and improve the video and input (DataChannel) stack.

Moreover, we have incorporated smaller frames for the Opus codec to see if the latency improves (tracked in #153), but NetEQ in Chrome mostly works on its own.

There are still multiple interventions that may bring this WebRTC stack to the maximum and achieve the most ideal and optimal performance possible.

**Backend:**

- Correctly implement YUV 4:4:4 color

https://issues.chromium.org/issues/40198264

This is possible in WebRTC, where Nutanix Frame implemented YUV 4:4:4 within Chromium quite some time ago.
First, however, color in YUV 4:2:0 (#160) should be solved first as there is no legitimate reason that color in YUV 4:2:0 should be over +/- 1 different from the original source.

- Obtain the sweet spot of video encoder maximum and minimum QP parameters

https://multi.app/blog/making-illegible-slow-webrtc-screenshare-legible-and-fast
https://multi.app/blog/measuring-shared-control-latency

- Investigate the usage of queues to GStreamer RTP payloaders

Currently, the Opus queue is commented out. However, queues may have useful features.
Along with re-investigating the effectiveness of queues in Opus and their roles in latency, queues in video RTP payloaders may (or may not) also help during congestion where certain latency spikes might stay for >5-15 seconds because the WebRTC decoder scrambles to decode very late frames instead of simply dropping them.
An unknown configuration from the web browser may also totally eliminate this situation.
This must work nicely with infinite keyframe/GOP configurations and NACK/PLI with RTX.

- Compress DataChannel using GZip

It seems that Nestri saw some effective input latency drops with this.

**Frontend:**

- Override system power settings (Especially Chromium on Windows) to decode full frames

https://web.dev/articles/requestvideoframecallback-rvfc
It seems that when the system is in a low power efficiency mode, video decoding is not done quickly, as in the example. This leads to perceived increased latency because the frames aren't getting painted as often as they should.
Some settings in WebRTC or <video> should prioritize full frames.

- ~~Moreover, `jitterBufferTarget` / `jitterBufferDelayHint` / `playoutDelayHint` are not well understood. Find out where this and other hidden WebRTC settings can improve upon the current approach.~~

**Update: DO NOT USE THIS! Another finding: `receiver.jitterBufferTarget = receiver.jitterBufferDelayHint = receiver.playoutDelayHint = 0`; is pretty bad to do it seems, causes stutter if the buffer wants to be bigger, but it's constantly forced down to 0 (most notably in higher resolutions).**

~~Current configuration (reference from https://groups.google.com/g/discuss-webrtc/c/wtuhQu6c1KY/m/Usq84y0mAQAJ, a bit of a CPU hog but acceptable with async, could be more optimized or otherwise able to assess the effect of this configuration in web browsers):~~

```javascript
// Repeatedly emit minimum latency target
webrtc.peerConnection.getReceivers().forEach((receiver) => {
    let intervalLoop = setInterval(async () => {
        if (receiver.track.readyState !== "live" || receiver.transport.state !== "connected") {
            clearInterval(intervalLoop);
            return;
        } else {
            receiver.jitterBufferTarget = receiver.jitterBufferDelayHint = receiver.playoutDelayHint = 0;
        }
    }, 15);
});
```

**WebRTC:**

**Check if merging `webrtcbin` back to one session is plausible: It seems that the `video-delay` could have reduced the video latency without needing to have two separate sessions.**

- Merge two different WebRTC sessions into one with multiple independent streams:

Use `a=rtcp-mux`, `a=group:BUNDLE 0 1 2 3 ...` and `a=mid:0`, `a=mid:1`, ... to establish one SDP session, but with independent streams for Audio, Video, DataChannel (`m=application x UDP/DTLS/SCTP webrtc-datachannel`), Microphone, Webcam, and other types of streams which don't interfere nor do audio/video sync.

Such as:

```
v=0
o=- 2 IN IP4 1.1.1.1
t=0 0
a=group:BUNDLE 0 1 2 3
a=fingerprint:sha-256
a=setup:actpass
m=audio x UDP/TLS/RTP/SAVPF 111 63
c=IN IP4 0.0.0.0
a=rtcp:x IN IP4 0.0.0.0
a=mid:0
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=sendonly
a=msid:id audio
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:111 opus/48000/2
a=rtcp-fb:111 transport-cc
a=fmtp:111 minptime=10;useinbandfec=1
a=rtpmap:63 red/48000/2
a=rtcp-fb:63 transport-cc
a=fmtp:63 111/111
a=ptime:10
m=video x UDP/TLS/RTP/SAVPF 96 97 101 102 98
c=IN IP4 0.0.0.0
a=rtcp:x IN IP4 0.0.0.0
a=mid:1
a=extmap:2 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:3 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:7 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:12 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=sendonly
a=msid:id video
a=rtcp-mux
a=rtcp-rsize
a=rtpmap:101 H264/90000
a=rtcp-fb:101 transport-cc
a=rtcp-fb:101 ccm fir
a=rtcp-fb:101 nack
a=rtcp-fb:101 nack pli
a=fmtp:101 level-asymmetry-allowed=1;packetization-mode=1;sps-pps-idr-in-keyframe=1;profile-level-id=42e01f
a=rtpmap:102 rtx/90000
a=fmtp:102 apt=101;rtx-time=125
m=application x UDP/DTLS/SCTP webrtc-datachannel
c=IN IP4 0.0.0.0
a=mid:2
a=sctp-port:5000
a=max-message-size:262144
m=audio x UDP/TLS/RTP/SAVPF 111
c=IN IP4 0.0.0.0
a=rtcp:x IN IP4 0.0.0.0
```

**The main purpose of doing this is to still isolate different streams so that there is no audio/video sync at all (otherwise adds inevitable latency) and at the same time improve the performance of DataChannels as well by maintaining an independent stream separate from the video, but handle all of them with one TURN relay port or other types of WebRTC port in one single SDP.**

- RTP Header Extensions and other WebRTC browser-side, server-side settings to implement and improve:

https://www.rtcbits.com/2023/05/webrtc-header-extensions.html

```
a=extmap:1 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:2 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
a=extmap:3 http://www.webrtc.org/experiments/rtp-hdrext/video-timing
a=extmap:4 http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
```

Note: `http://www.webrtc.org/experiments/rtp-hdrext/color-space` causes the Chrome WebRTC decoder to skip the Hardware Decoder and go straight to the Software FFmpeg decoder.

The above RTP Header Extensions are known to help with controlling latency and timing. These can be implemented in GStreamer so that it can be emitted into RTP payloaders.

https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/3549
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/issues/3550

https://gstreamer.freedesktop.org/documentation/rtpmanager/rtphdrextclientaudiolevel.html
https://gstreamer.freedesktop.org/documentation/rtpmanager/rtphdrextmid.html

SDP support in web browsers: https://codepen.io/kwst/full/yLaaxRy

`draft-holmer-rmcat-transport-wide-cc-extensions-01` is enabled for video and audio when `rtpgccbwe` is active. `abs-send-time`, `video-timing` are not available in GStreamer. `playout-delay` has been implemented in a very restricted temporary form in `gstwebrtc_app.py`, where the only zero values can be sent (which is what we need, anyways).

- Investigate `imageattr` and `flexfec` in video:

```
a=imageattr:96 send [x=[1280:1920],y=[720:1080],fps=[30:60]]
a=imageattr:97 send [x=[1280:1920],y=[720:1080],fps=[30:60]]
a=rtpmap:98 flexfec-03/90000
a=rtcp-fb:98 transport-cc
a=fmtp:98 repair-window=10000000
a=ssrc-group:FEC-FR
```

- Larger DataChannels:

```
a=max-message-size:262144
```

- Understand the effects of `b=AS:` and `x-google-max-bitrate` (in the receiving-side, not the sending-side or browser-to-browser !!):

https://github.com/nextcloud/spreed/issues/6739
https://groups.google.com/g/discuss-webrtc/c/u7k1_hASS4Q
https://stackoverflow.com/questions/57653899/how-to-increase-the-bitrate-of-webrtc
https://groups.google.com/g/discuss-webrtc/c/udyHHPnrQMo
https://github.com/pion/webrtc/discussions/1827
https://ekobit.com/blog/diving-deeper-into-webrtc-advanced-options-and-possibilities/
https://chromium.googlesource.com/external/webrtc/+/a6b99448eec51527eca0bc59f6da71061d02e807/webrtc/media/base/mediaconstants.cc
https://groups.google.com/g/discuss-webrtc/c/ORJdeoFAaBE
https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/sdp-ext/fmtp-x-google-per-layer-pli.md

The above links may have irrelevant information (controlling sender bitrate, this is because `webrtcbin` is the sender and it does not use libwebrtc).

```
b=AS:300000
a=fmtp:96 sps-pps-idr-in-keyframe=1;x-google-max-bitrate=300000;x-google-min-bitrate=0;x-google-start-bitrate=12000
```

- Different protocol topologies to TURN and STUN (such as ICE-TCP)

https://neko.m1k1o.net/#/getting-started/configuration?id=webrtc

Pion provides various WebRTC configurations and protocols including EPR, UDPMUX, TCPMUX, NAT1TO1, ICE-LITE, ICE-TCP, etc. These techniques allow more setup flexibility in addition to TURN/STUN and allow limiting port ranges or using a single port for many numbers of connections. This should be implemented with GStreamer's `webrtcbin`.

https://www.w3.org/2021/03/media-production-workshop/talks/slides/sergio-garcia-murillo-whip.pdf
https://groups.google.com/g/discuss-webrtc/c/wtuhQu6c1KY
https://henbos.github.io/webrtc-timing/
https://github.com/jakearchibald/web-platform-tests/blob/master/webrtc-extensions/RTCRtpReceiver-playoutDelayHint.html
https://mediasoup.discourse.group/t/webrtc-playout-delay-extension/2067
https://issues.chromium.org/issues/324276557
https://bugzilla.mozilla.org/show_bug.cgi?id=1592988
https://groups.google.com/a/chromium.org/g/blink-dev/c/4W4orKqA3Rs
https://www.reddit.com/r/WebRTC/comments/ipewaq/disable_use_of_jitter_buffer/?rdt=58693

### LINKED ISSUE #200: Xemu fails to detect remote gamepad
﻿
﻿
**Describe the bug**
﻿
The Xemu emulator, when running inside a Lutris container, fails to detect a connected gamepad (Switch Pro Controller). The controller is recognized by other applications and emulators (Wine, PCSX3, Yuzu) within the same container, but not by Xemu. The issue occurs when the controller is connected to a client machine accessing the Lutris host over the local network via a browser.
﻿
1. Connect a Switch Pro Controller to a client computer.
2. On the client computer, open a web browser and connect to the Lutris host machine on the local network.
3. Launch the Xemu emulator from within the Lutris interface.
4. Attempt to configure or use the controller in Xemu's input settings.
5. Observe that the controller is not detected by Xemu.
﻿
**Expected behavior**
﻿
Xemu, when running within the Lutris container, should detect and allow the use of the Switch Pro Controller that is connected to the remote client, consistent with the behavior of other emulators like PCSX3 and Yuzu in the same environment.
﻿
**Screenshots**

Image

- **Host OS Version**: Ubuntu 24.04
- **Browser Version**: [Chrome 139.0.7258.139]
- **Controller Model**: Nintendo Switch Pro Controller
﻿
**Additional context**
﻿
* The fix from commit #173 has been applied.
* The controller works as expected in Wine, PCSX3, and Yuzu within the same new container.
* The issue has been tested and confirmed with both Xemu v0.7.111 and the latest version, v0.8.96.
* The controller is detected correctly by Xemu if it is physically plugged into the host machine. The problem only occurs when the controller is connected to a remote client accessing the host via a browser.

 - [x] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
 - [x] I confirm that I have read other open and closed issues and that duplicates do not exist.
 - [x] I confirm that the issue is easily reproducible and explained thoroughly.
 - [x] I confirm that relevant log files have been included as explained below. Any relevant additional log files have also been included.
 - [x] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [x] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

### LINKED ISSUE #214: After typing the @ in an italian keyboard layout, all the following letters I type are in uppercase
**Describe the bug**
After I type the @ in the italian keyboard layout (in PC: ALT GR + ò, in Mac: OPTION + ò ) all the following letters I type are in uppercase, it would seem like the SHIFT key remains pressed, in fact after pressing SHIFT once all returns to normal. 

**To Reproduce**
1. Add an Italian keyboard layout to your HOST system (if you don't already have one)
2. Select that Layout 
3. open the selkies image
4. open XTerm
5. Type the @ - [ALT GR + ò] or [OPTION + ò]
6. Type other stuff (see the uppercases)
7. Type SHIFT and see how everything returns to normal
 
**Expected behavior**
Typing the @ in any keyboard layout should work the same without producing the effect explained previously

**Screenshots**

Image

https://github.com/user-attachments/assets/cef74e19-6d43-4cee-9648-3d6cdf0c923a

 - Host OS Version: MacOS 26.2
 - Host GPU Model and Driver/Encoder Version: Apple M2 Pro
 - GStreamer Version: not relevant
 - Browser Version : Chrome 143.0.7499.193
 - base image used: ghcr.io/linuxserver/baseimage-selkies:ubuntunoble    VERSION: b5539e10-ls130

**Additional context**

 - [X] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
 - [X] I confirm that I have read other open and closed issues and that duplicates do not exist.
 - [X] I confirm that the issue is easily reproducible and explained thoroughly.
 - [X] I confirm that relevant log files have been included as explained below. Any relevant additional log files have also been included.
 - [X] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [X] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

### LINKED ISSUE #220: [BUG] Cursor rendering lost on desktop browsers after tab sleep - restored only by mobile client reconnect (Wayland/NVIDIA)
## Current Behavior

When using the webtop in Wayland mode (`PIXELFLUX_WAYLAND=true`) with NVIDIA GPU passthrough, the mouse cursor becomes invisible on all desktop browsers after the browser tab goes to sleep or is backgrounded. Cursor **input** continues to function correctly (clicks, hover states, interactions all work), but the cursor is not rendered visually.

The issue persists across:
- Multiple desktop browsers (Microsoft Edge, Firefox)
- Incognito/private windows
- Changing stream resolution or quality settings in the Selkies sidebar
- Opening a new browser session on the same desktop

The cursor is **only** restored by connecting from a **mobile browser** (tested: Microsoft Edge on iOS). Once the mobile client connects and disconnects, the cursor renders correctly again on all desktop browsers.

The cursor also will render and function normally while in 'Gaming Mode'. This acts as a work around, but this does not restore the cursor after exiting Gaming Mode. 

## Expected Behavior

The mouse cursor should remain visible after a browser tab goes to sleep or loses focus, and should be restorable without requiring a mobile client connection or full container restart.

## Steps to Reproduce

1. ~~Launch the webtop container in Wayland mode (`PIXELFLUX_WAYLAND=true`) with NVIDIA GPU~~
2. Launch webtop w
3. Connect from a desktop browser (Edge or Firefox)
4. Allow the browser tab to go to sleep / background the tab for an extended period
5. Return to the tab — cursor is invisible but input still functions
6. Open the webtop URL in a mobile browser (e.g. Edge on iOS)
7. Observe cursor is restored on the mobile session
8. Switch back to the desktop browser — cursor now renders correctly, but the cursor appearance differs from the original (e.g. changes cursor theme or type (cursor and not a pointer)), suggesting the mobile connection triggers a cursor reinitialization rather than restoring the previous cursor state. 
## Environment

- **Image:** `lscr.io/linuxserver/webtop:ubuntu-kde` (custom Dockerfile build)
- **Host OS:** Ubuntu 6.8.0-101 x86_64
- **Docker version:** Docker version 29.2.1
- **GPU:** NVIDIA (passthrough via `deploy.resources.reservations.devices`)
- **Desktop Environment:** KDE Plasma (Wayland or CPU Only)
- **Streaming backend:** Selkies (`svc-selkies`)
- **Client browsers affected:** Microsoft Edge (desktop), Firefox (desktop)
- **Client browser NOT affected:** Microsoft Edge (iOS)

### Wayland
```
PIXELFLUX_WAYLAND=true
DRINODE=/dev/dri/renderD128
DRI_NODE=/dev/dri/renderD128
NVIDIA_VISIBLE_DEVICES=all
NVIDIA_DRIVER_CAPABILITIES=all
WLR_NO_HARDWARE_CURSORS=1
```
### CPU 
```
PIXELFLUX_WAYLAND=true
WLR_NO_HARDWARE_CURSORS=1
WLR_RENDERER=pixman
```

## Additional Context

- `WLR_NO_HARDWARE_CURSORS=1` no change
- Enabling CSS cursors via the Selkies sidebar (`use_browser_cursors`) actually triggers the issue immediately rather than fixing it
- Restarting the container restores the cursor, suggesting Selkies session state is not fully reset on reconnect
- The fact that a mobile client connection restores cursor rendering for all subsequent desktop clients suggests the mobile connection triggers a cursor stream reinitialization path that desktop reconnects do not
- Cursor input (clicks, hover states) continues to function correctly throughout — this is purely a rendering issue

### LINKED ISSUE #222: [BUG] Left Ctrl/Alt keys do not work when swapped using xkb on host
**Describe the bug**

When I try to bold or italicize using Obsidian in a Selkies instance, it does not work with Ctrl/Alt keys that have been rebound. The keys instead revert to the browser, which prevents them from being used at all.

I am having this issue with multiple apps and multiple browsers, and have seen it in both Niri and GNOME. The issue appears tied to how my keys are bound in the system, not anything underlying in the system itself. In fact, reversing the binding fixes the issue.

My xkb configuration uses variant "mac" (necessary for third-level keys like long dashes and curly quotes) which may be triggering a Mac-specific code path in Selkies that remaps AltLeft to XK_Meta_L. This shim should either not apply on Linux regardless of keyboard variant, or should be opt-in rather than automatically triggered by keyboard layout detection.

**To Reproduce**

Using a Linux distro, modify the key bindings using xkb (switch LCtrl with LAlt should be enough). My approach, for testing:

```
layout "us"
variant "mac"
options "lv3:lwin_switch,ctrl:swap_lalt_lctl,caps:super"
```

Install Obsidian, Chromium, or similar apps using a server-based tool like Unraid, and visit it in browser. Attempt to use a command using a Ctrl or Alt key, such as copy, paste or undo.

**Expected behavior**

When pressing the Left Alt key, it should work like Left Ctrl. Instead, it is a dead key. However, the Left Alt does not work either. Both instead favor the browser. The Right Ctrl and Alt keys, which are not modified, do not have this issue, and when the xkb settings are disabled, it works correctly.

(The third level key in my setup continues to work.)

The result is that:

1. Browser sends AltLeft to Selkies
2. Selkies bug maps AltLeft to XK_Meta_L
3. Obsidian receives Meta instead of Ctrl
4. Nothing works

**Screenshots**

 - Host OS Version: Bazzite DX 43
 - Host GPU Model and Driver/Encoder Version: Intel ARC A370M, Intel Iris XE
 - GStreamer Version: GStreamer 1.26.10
 - Browser Version: Vivaldi 7.8.3925.81 (tested on other browsers)
 - Any other information specific to your environment: Usually run niri but also have confirmed the issue in GNOME as well.

**Additional context**

I have tested game mode to see if it does not have an effect on this bug, and it does not.

 - [x] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
 - [x] I confirm that I have read other open and closed issues and that duplicates do not exist.
 - [x] I confirm that the issue is easily reproducible and explained thoroughly.
 - [x] I confirm that relevant log files have been included as explained below. Any relevant additional log files have also been included.
 - [x] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [x] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

### LINKED ISSUE #234: Adjust clipboard handling to make seamless keyboard work on Safari & Firefox
**Is your enhancement request related to a problem? Please describe.**
On Safari & Firefox, pasting to and copying from server does not work. Even though the Web standards currently allow for this.

**Describe the solution you'd like**

**Paste to Server:** Move clipboard read to a keydown handler (Crtl + V) instead.

**Copy from Server:** Move clipboard write to a keydown handler (Crtl + C) instead of on window-focus, and pass a Promise for the data, to allow waiting for data from the server.

From: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

> Firefox & Safari:
>
> - If a read isn't allowed by the spec but transient user activation is still met, it triggers a user prompt in the form of an ephemeral context menu with a single "Paste" option (which becomes enabled after 1 second) and succeeds if the user chooses the option.
> - Writing requires transient activation.
> - The paste-prompt is suppressed if reading same-origin clipboard content, but not cross-origin content.
> - The clipboard-read and clipboard-write permissions are not supported (and not planned to be supported) by Firefox or Safari.

**Describe alternative implementation you have seen**

Implement `document.execCommand('copy')` fallback strategy as mentioned here: https://medium.com/@hailieejkim/safari-clipboard-errors-notallowederror-and-typeerror-fixes-e9c23becd856

See example code: https://stackoverflow.com/a/74528564

**Additional context**

 - [x] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
 - [x] I confirm that I have read other open and closed issues and that duplicates do not exist.
 - [x] I confirm that I have described the solution as well as the problem in detail, and if possible, explained how to solve the problem.
 - [x] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [x] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

### LINKED ISSUE #241: [BUG] Consecutive mouse wheel events are dropped or coalesced, causing slow scrolling
**Describe the bug**

When scrolling quickly with a mouse wheel in a LinuxServer.io Webtop container using Selkies, consecutive wheel events appear to be dropped or coalesced before they reach the remote X11 session.

As a result, scrolling feels much slower than expected. In the reproduction below, spinning the mouse wheel quickly over an `xev` window produces only a single X11 scroll event (`button 5`) in the container.

This was originally noticed in `linuxserver/docker-obsidian`, but the same behavior is reproducible in `linuxserver/webtop:latest`, so it does not appear to be specific to Obsidian or Electron.

**To Reproduce**

1. Run a LinuxServer.io Webtop container:

```bash
docker run --rm -it \
  --name webtop-selkies-test \
  --shm-size=1gb \
  --security-opt seccomp=unconfined \
  -e PUID=1000 \
  -e PGID=1000 \
  -e TZ=Asia/Tokyo \
  -p 3000:3000 \
  -p 3001:3001 \
  lscr.io/linuxserver/webtop:latest
````

2. Open the Webtop session in the browser.

3. Install `xev` inside the container:

```bash
docker exec -it webtop-selkies-test bash
sudo apk add --no-cache xev
# or, if already running as root:
apk add --no-cache xev
```

4. Run `xev` with button events only:

```bash
xev -event button
```

5. Move the mouse cursor over the `xev` window.

6. Spin the mouse wheel quickly.

7. Observe that only one scroll event is reported, for example:

```text
root@c3626a3c5cca:/# xev -event button
Outer window is 0x2200001, inner window is 0x2200002

ButtonPress event, serial 25, synthetic NO, window 0x2200001,
    root 0x21f, subw 0x0, time 775420882, (61,78), root:(1232,659),
    state 0x0, button 5, same_screen YES

ButtonRelease event, serial 25, synthetic NO, window 0x2200001,
    root 0x21f, subw 0x0, time 775420882, (61,78), root:(1232,659),
    state 0x1000, button 5, same_screen YES
```

**Expected behavior**

When the mouse wheel is spun quickly, multiple wheel events should either reach the remote X11 session as multiple `ButtonPress` / `ButtonRelease` events, or the accumulated wheel delta should be preserved in a way that produces equivalent scrolling behavior.

For example, a fast downward wheel movement should produce multiple `button 5` events in `xev`, not only a single press/release pair.

**Screenshots**

No screenshot. The issue is observable through the `xev -event button` output above.

* Host OS Version: Windows 11 with WSL2

  * WSL Version: 2.6.3.0
  * Kernel Version: 6.6.87.2-1
  * WSLg Version: 1.0.71
  * MSRDC Version: 1.2.6353
  * Direct3D Version: 1.611.1-81528511
  * DXCore Version: 10.0.26100.1-240331-1435.ge-release
  * Windows Version: 10.0.26200.8246
  * WSL distribution: Ubuntu 24.04.3 LTS (noble)
* CPU: 13th Gen Intel(R) Core(TM) i5-13600KF
*  Host GPU Model and Driver/Encoder Version: NVIDIA GeForce RTX 4070 Ti, NVIDIA driver 595.97, CUDA 13.2
*  GStreamer Version: not available in this container image; `gst-launch-1.0` and `gst-inspect-1.0` are not installed
*  Browser Version: Brave 1.89.141 (Chromium 147.0.7727.102)
*  Docker Version: Docker version 29.2.1, build a5c7197
*  Container image: `lscr.io/linuxserver/webtop:latest`
* 

 Container image ID: `sha256:a7bc86b5d2e8cfb03af0d1eafa95253498f4ee413a70e63f262190fa1db96d8a`

  * The behavior was first noticed in `linuxserver/docker-obsidian`.
  * It was reproduced in `linuxserver/webtop:latest` to isolate the problem from Obsidian/Electron.
  * Logs are long and will be attached separately as a  [log.txt](https://github.com/user-attachments/files/27287930/log.txt) file.

**Additional context**

This looks different from the existing touchpad issues where scrolling is too fast. In this case, rapid wheel input appears to result in too few X11 scroll events, making scrolling feel slow.

The main observation is that continuous physical mouse wheel input produces only a single `button 5` press/release pair in the remote X11 session during a fast wheel movement.

* [x] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
* [x] I confirm that I have read other open and closed issues and that duplicates do not exist.
* [x] I confirm that the issue is easily reproducible and explained thoroughly.
* [x] I confirm that relevant log files have been included as explained below. Any relevant additional log files have also been included.
* [x] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
* [x] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

```
```

### LINKED ISSUE #257: Letters entered via keepassxc auto-type are all lowercase
It is possible to use [keepassxc](https://github.com/keepassxreboot/keepassxc) [auto-type](https://keepassxc.org/docs/KeePassXC_UserGuide#_auto_type) to enter the username and password etc. into the selkies container App in the browser, but the entered content will all be converted to lowercase, which is not work for passwords most of the time.

I'm not sure why this happens, since even typing in uppercase letters via the virtual on-screen keyboard works.

Another related issue is that if the inter-key delay of auto-type is too small (e.g. less than 10ms), individual letters may be lost resulting in input errors. This doesn't happen in other web pages in the browser, but it does in selkies.

In fact, upon closer inspection, all that seems to be missing are special characters such as `_` or `!`, I guess this has something to do with the handle of the modifier keys, and it may be the same reason that caused the first issue.

### LINKED ISSUE #258: Interference in reading the clipboard in Firefox
I'm not sure about other browsers, but in `Firefox`, due to reading the system clipboard triggers a `Paste` button that has a one-second delay before it can be clicked, which is usually annoying and a blocking in many situations.

Switching back to the browser from other windows often triggers this button, leading to:
- Regardless of whether the button is clicked or not, it usually takes 4-5 clicks to generate a truly effective click.
- When using keepassxc for auto-type, input is blocked and canceled due to the `Paste` button trigger.

Normally, pasting text into an input-text or textarea element does not trigger the `Paste` button, but in selkies users are unable to paste wanted content when using shortcuts such as `Ctrl/Command + v`, since is not the time for selkies to synchronize clipboard.

These experiences lead to a completely opposite and annoying experience: I am frequently interrupted when I don't want to paste from the clipboard, and it doesn't work when I DO wanna paste from the clipboard.

I even tried using only the selkies sidebar's clipboard textarea and preventing it with `SELKIES_CLIPBOARD_IN_ENABLED=False`, but that didn't work because it only prevented the content from being sent to the server, but the `Paste` button still appeared in Firefox, in other words, JS is still trying to read the clipboard.

Most of the time I just want to paste some text, and I haven't enabled image or binary support. But the current clipboard input experience in Firefox is terrible.

Some relevant materials:
- https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations
- https://bugzilla.mozilla.org/show_bug.cgi?id=1774163
- https://groups.google.com/g/novnc/c/UsrrYQS-Agw

I think some possible improvements or solutions are:
- Avoid triggering the `Paste` button by only reading from a textarea when image and binary transfers are disabled.
- Stop using JS to read the clipboard when `SELKIES_CLIPBOARD_IN_ENABLED=False`.
- Use a Firefox extension to read the clipboard, https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard#reading_from_the_clipboard

### LINKED ISSUE #259: Environment changes were not sync to the client in time
In #258, I tried using `SELKIES_CLIPBOARD_IN_ENABLED=False` to work around the bad experience, but I found that wasn't what I expected, and it actually prevented the sidebar clipboard from working as well.

So I unset that environment variable, but strangely, no matter how many times I restarted the container, tab, and browser, it still wouldn't back to work.

I eventually discovered that the variable was still marked as False in the tab's local storage, and it only started working again after I manually cleared the local storage.

Therefore, it seems that changes to container environment variables are not promptly synchronized to the browser client.

ehfd:
Must be tested in tandem to https://github.com/linuxserver/pixelflux/pull/18 and 
https://github.com/linuxserver/pcmflux/pull/5. Interface and API/ABI are backward-incompatible.

Fixes #277 
Fixes #276
Fixes #275
Fixes #273
Fixes #271
Fixes #270
Fixes #269
Fixes #268
Fixes #267
Fixes #266 
Fixes #265 
Fixes #264 
Fixes #261 
Fixes #259 
Fixes #258 
Fixes #257 
Fixes #245
Fixes #241 
Fixes #234
Fixes #222 
Fixes #220 
Fixes #214 
Fixes #200
Fixes #169
Fixes #157
Fixes #153
Fixes #145
Fixes #85


rmounce:
Verified the incorporation of our #264/#265/#266/#267 at 0bf1639 — all four look correctly integrated (and the adaptive paint cushion with `selkiesVideoStats` diagnostics is a nicer design than either of the originals; we'll exercise it in testing).

One small gap spotted while checking the storage-keying integration: `addons/selkies-dashboard-wish/src/components/dashboard/settings.tsx` (`getStorageAppName()`, ~line 36) adopted the corrected sanitizer regex but still derives from `window.location.href.split('#')[0]` rather than `window.location.origin + window.location.pathname`. Its comment says it must match the streaming cores' prefix, but with the query string still included, a tokenized deployment gets (a) token-scoped keys leaking from the wish dashboard and (b) the wish dashboard reading/writing a different namespace than the cores on the same page. The main `selkies-dashboard` Sidebar has the corrected derivation, so it's just this one site.

We're building the test image for this branch (+ linuxserver/pixelflux#18, linuxserver/pcmflux#5) against our production NVIDIA setup now; findings to follow here.

ehfd:
I am also screening for more issues during the rewrite. I'll update when I'm finished.

ehfd:
@rmounce I fixed the issue you stated, as well as finishing off other things.

Note that there may be basic errors since this is fresh code, but it should be trivial to fix.

Feel free to PR anything to either of the three branches.

rmounce:
**Updated 2026-07-13:** SHAs bumped to the current branch tips and the nginx rule below reworked, because the WebSocket endpoints also moved under `/api` in the recent frontend/server rework (`{prefix}/api/websockets` for the data plane, `{prefix}/api/ws` for WebRTC signaling — nothing the browser dials is outside `/api` anymore). The previous revision of this recipe fails against current tips with a browser-side "can't establish a connection to wss://.../api/websockets" because its plain `/api` proxy location stripped the upgrade headers; there are now dedicated upgrade locations per ws path, and the plain `/api` location raises `client_max_body_size` for the new 64 MiB file-upload POST slices. Anything else in an image that dials the old paths must follow the move too (e.g. a token helper on `{prefix}/websockets` → `{prefix}/api/websockets`). Also new: the `uvloop` optional dependency (the SCTP data-channel throughput improvement relies on it).

**Updated 2026-07-08:** pixelflux no longer needs source-built ffmpeg/x264/libjpeg-turbo/wayland (system ffmpeg 6.0–8.1 is accepted, libjpeg-turbo is vendored statically, wayland is dlopen'd) — the builder stage is now just distro dev packages + rustup.

For anyone else wanting to second-eye test this stack on the linuxserver.io images: below is the Dockerfile we've been using, sanitised to build directly on `ghcr.io/linuxserver/baseimage-selkies:ubuntunoble`. It builds the pixelflux/pcmflux wheels from the PR branches (linuxserver/pixelflux#18, linuxserver/pcmflux#5), the web frontend and the selkies package from this branch, and patches them into the stock image in place, working around the noble-specific gaps we reported earlier (the `six`/`uvloop` runtime deps, the LSIO-required `nginx/` fancyindex files, the `/api` proxy locations, and websocket basic auth vs browser clients).

The pins are what we've actually exercised together; bump them as a set.

<details><summary>Dockerfile</summary>

```dockerfile
# Example: build and test the comprehensive-fixes stack on top of the
# linuxserver.io Selkies base image, patching the stock image in place:
#
#   - selkies-project/selkies#254   (comprehensive-fixes: selkies python + frontend)
#   - linuxserver/pixelflux#18      (comprehensive-fixes: pure-Rust rewrite)
#   - linuxserver/pcmflux#5         (comprehensive-fixes: pure-Rust, AudioPlayback)
#
# This is a sanitised copy of the recipe we use for second-eye testing #254 —
# our real builds start from a downstream image based on
# baseimage-selkies:ubuntunoble, which is what this stack has actually been
# exercised on (Intel/VA-API and NVIDIA/NVENC, Firefox + Chrome clients).
#
# Since the first version of this example, pixelflux stopped pinning ffmpeg
# (any system 6.0-8.1 works, only avcodec/avfilter needed), statically
# vendors libjpeg-turbo, links noble's libx264-dev, and dlopens wayland — so
# the x264/libjpeg-turbo/ffmpeg/wayland source builds are all gone in favour
# of noble's distro packages, and the builder stage is just deps + rustup.
#
# All three source branches are moving targets that advance by amend +
# force-push; the SHAs below pin what we tested. Bump them together — the
# selkies python side needs the matching native-module features
# (deferred_free / OwnedFrame / request_idr_frame / AudioPlayback).
ARG SELKIES_REPO=https://github.com/selkies-project/selkies
ARG SELKIES_HASH=bf28eadc1669aebcafc8728dfb1c92a00b03812b
ARG PIXELFLUX_HASH=20d5254cf065c556d5c53d2f098212a9fea9ae84
ARG PCMFLUX_HASH=e78167490b2fbb1535f908143e44834b8f3e2292
ARG BASE_IMAGE=ghcr.io/linuxserver/baseimage-selkies:ubuntunoble

########################################################################
# Frontend: mirrors the base image's own packaging flow.
########################################################################
FROM node:20-alpine AS frontend
RUN apk add --no-cache git
ARG SELKIES_REPO
ARG SELKIES_HASH
RUN mkdir /src && cd /src && \
        git init && git remote add origin "${SELKIES_REPO}.git" && \
        git fetch --depth 1 origin "${SELKIES_HASH}" && git checkout FETCH_HEAD

# The postbuild step (gendb.js) fetches the SDL game-controller DB from
# raw.githubusercontent.com and swallows fetch failures (logs + exit 0),
# silently skipping dist/jsdb — which the dashboard stage's cp then trips
# over. Retry it once and fail loudly here instead.
RUN cd /src/addons/selkies-web-core && \
        npm install && npm run build && \
        { [ -d dist/jsdb ] || node gendb.js; } && \
        { [ -d dist/jsdb ] || { echo 'ERROR: dist/jsdb missing - gendb.js could not fetch SDL_GameControllerDB (transient network?). Re-run the build.'; exit 1; }; }
# The final stage keeps the base image's nginx/ fancyindex header/footer
# copies (required by the LSIO runtime: init-nginx seds footer.html and
# default.conf references both), so they are not packaged from source here.
RUN cd /src/addons/selkies-dashboard && \
        cp ../selkies-web-core/dist/selkies-core.js src/ && \
        npm install && npm run build && \
        mkdir -p dist/src && \
        cp ../selkies-web-core/dist/selkies-core.js dist/src/ && \
        cp ../universal-touch-gamepad/universalTouchGamepad.js dist/src/ && \
        cp -r ../selkies-web-core/dist/jsdb dist/

########################################################################
# Native modules: build pixelflux/pcmflux wheels from the PR branches.
#
# Both are pure-Rust PyO3 extensions (no prebuilt wheels for these
# branches). Building FROM the same base image guarantees the wheels match
# /lsiopy's python and the base glibc.
#
# The dep list mirrors pixelflux's [tool.cibuildwheel] before-all:
# ffmpeg-sys-next accepts system ffmpeg 6.0-8.1 (noble: 6.1) with only
# avcodec/avfilter needed; x264-sys links noble's libx264-dev; turbojpeg-sys
# statically builds its vendored libjpeg-turbo (hence cmake + nasm); wayland
# is dlopen'd at runtime, so headers suffice.
########################################################################
FROM ${BASE_IMAGE} AS native-builder
ENV DEBIAN_FRONTEND=noninteractive
RUN apt-get update && apt-get install -y --no-install-recommends \
        build-essential clang libclang-dev curl git nasm cmake pkg-config \
        libpixman-1-dev libx11-dev libxext-dev libxfixes-dev \
        libva-dev libdrm-dev libinput-dev libxkbcommon-dev libgbm-dev \
        libwayland-dev libudev-dev \
        libavcodec-dev libavfilter-dev libx264-dev \
        libpulse-dev libopus-dev

# noble's rustc is too old for the PyO3 0.27 crate graph. The LSIO base image
# sets HOME=/config, so don't let rustup pick $HOME/.cargo — pin both trees
# somewhere fixed and PATH that instead.
ENV RUSTUP_HOME=/opt/rustup \
    CARGO_HOME=/opt/cargo \
    PATH=/opt/cargo/bin:$PATH
# Download-then-run, NOT `curl | sh`: a failed download would feed sh an empty
# script, which exits 0 and silently installs nothing.
RUN curl --proto '=https' --tlsv1.2 -sSf -o /tmp/rustup-init.sh https://sh.rustup.rs && \
        sh /tmp/rustup-init.sh -y --profile minimal --no-modify-path && \
        rm /tmp/rustup-init.sh
RUN cargo --version && rustc --version

# The PR heads live on ehfd's forks (linuxserver/pixelflux#18,
# linuxserver/pcmflux#5); init+fetch pins an arbitrary SHA on any git.
ARG PIXELFLUX_HASH
ARG PCMFLUX_HASH
RUN mkdir /tmp/pixelflux && cd /tmp/pixelflux && \
        git init && git remote add origin https://github.com/ehfd/pixelflux.git && \
        git fetch --depth 1 origin "${PIXELFLUX_HASH}" && git checkout FETCH_HEAD && \
        pip wheel --no-deps -w /wheels .
RUN mkdir /tmp/pcmflux && cd /tmp/pcmflux && \
        git init && git remote add origin https://github.com/ehfd/pcmflux.git && \
        git fetch --depth 1 origin "${PCMFLUX_HASH}" && git checkout FETCH_HEAD && \
        pip wheel --no-deps -w /wheels .

# The joystick interposer and fake-udev shim were both reworked on
# comprehensive-fixes; the base image's copies predate that, so rebuild them
# from the same selkies checkout the python side comes from.
ARG SELKIES_REPO
ARG SELKIES_HASH
RUN mkdir /tmp/selkies && cd /tmp/selkies && \
        git init && git remote add origin "${SELKIES_REPO}.git" && \
        git fetch --depth 1 origin "${SELKIES_HASH}" && git checkout FETCH_HEAD && \
        cd addons/js-interposer && \
        gcc -shared -fPIC -o selkies_joystick_interposer.so joystick_interposer.c -ldl -pthread && \
        cd ../fake-udev && \
        make

########################################################################
# Final: patch the base image in place.
########################################################################
FROM ${BASE_IMAGE}
ARG SELKIES_REPO
ARG SELKIES_HASH
ENV DEBIAN_FRONTEND=noninteractive

# Runtime halves of the distro libs the wheels link against (noble ffmpeg 6.1
# sonames + x264; libjpeg-turbo is statically vendored inside the pixelflux
# wheel). NVIDIA driver libs (libcuda, libnvidia-encode) come from the
# container runtime as usual — the Rust NVENC bindings dlopen them.
RUN apt-get update && apt-get install -y --no-install-recommends \
        libavcodec60 libavfilter9 libx264-164 libopus0 && \
    rm -rf /var/lib/apt/lists/*

COPY --from=native-builder /wheels /tmp/wheels
RUN pip install --no-deps --force-reinstall /tmp/wheels/*.whl && rm -rf /tmp/wheels

# selkies python package. 'six' is a runtime dep on comprehensive-fixes (its
# vendored python-Xlib needs it) and 'uvloop' is an optional one (__main__
# falls back to the stock loop, but the SCTP data-channel throughput fix
# depends on it), so install both explicitly. (pynput / GPUtil / pasimple /
# python-xlib were dropped upstream; the stale copies in the image are
# harmless.)
RUN pip install --no-deps --force-reinstall \
        "${SELKIES_REPO}/archive/${SELKIES_HASH}.zip" && \
        pip install six uvloop && \
        rm -rf /tmp/* ~/.cache

# Replace the whole dashboard build (content-hashed assets must stay
# consistent), but keep the base image's nginx/ fancyindex header/footer,
# which the LSIO runtime requires (init-nginx seds footer.html; default.conf
# references both).
RUN cp -a /usr/share/selkies/selkies-dashboard/nginx /tmp/nginx-keep && \
        rm -rf /usr/share/selkies/selkies-dashboard
COPY --from=frontend /src/addons/selkies-dashboard/dist/ \
        /usr/share/selkies/selkies-dashboard/
RUN mkdir -p /usr/share/selkies/selkies-dashboard/nginx && \
        cp -a /tmp/nginx-keep/. /usr/share/selkies/selkies-dashboard/nginx/ && \
        rm -rf /tmp/nginx-keep

# The LSIO nginx only proxies SUBFOLDERwebsocket* to the backend; on
# comprehensive-fixes EVERYTHING dynamic lives under SUBFOLDERapi/ — the
# control plane (status, health, tokens, switch, turn, upload/files) AND the
# WebSocket endpoints themselves (SUBFOLDERapi/websockets data plane,
# SUBFOLDERapi/ws WebRTC signaling), so the base image's SUBFOLDERwebsocket
# location no longer matches anything the client dials. nginx maps can't
# live in default.conf (server context), so each ws path gets an upgrade
# location cloned from the base websocket block — note "websockets" does
# NOT start with "ws" as a string, so the two need separate prefix
# locations — plus a plain SUBFOLDERapi for the HTTP endpoints with
# client_max_body_size 65M (file uploads POST Blob slices of up to 64 MiB
# to SUBFOLDERapi/upload; nginx's 1M default would 413 them). The SUBFOLDER
# and CWS placeholders are substituted by the image's init scripts at
# container start.
#
# NOTE: anything in the image that talks to the old endpoints must follow
# this move too — e.g. a token-provisioning helper POSTing to /tokens must
# now POST to {prefix}/api/tokens, a native client dialing
# {prefix}/websockets must dial {prefix}/api/websockets, and until the
# first successful tokens POST the server's config gate holds all websocket
# connections.
RUN sed -i "s|location SUBFOLDERwebsocket {|\
location SUBFOLDERapi/websockets {\n\
    proxy_set_header        Upgrade \$http_upgrade;\n\
    proxy_set_header        Connection \"upgrade\";\n\
    proxy_set_header        Host \$host;\n\
    proxy_set_header        X-Real-IP \$remote_addr;\n\
    proxy_set_header        X-Forwarded-For \$proxy_add_x_forwarded_for;\n\
    proxy_set_header        X-Forwarded-Proto \$scheme;\n\
    proxy_http_version      1.1;\n\
    proxy_read_timeout      3600s;\n\
    proxy_send_timeout      3600s;\n\
    proxy_connect_timeout   3600s;\n\
    proxy_buffering         off;\n\
    client_max_body_size    10M;\n\
    proxy_pass              http://127.0.0.1:CWS;\n\
  }\n\n\
  location SUBFOLDERapi/ws {\n\
    proxy_set_header        Upgrade \$http_upgrade;\n\
    proxy_set_header        Connection \"upgrade\";\n\
    proxy_set_header        Host \$host;\n\
    proxy_set_header        X-Real-IP \$remote_addr;\n\
    proxy_set_header        X-Forwarded-For \$proxy_add_x_forwarded_for;\n\
    proxy_set_header        X-Forwarded-Proto \$scheme;\n\
    proxy_http_version      1.1;\n\
    proxy_read_timeout      3600s;\n\
    proxy_send_timeout      3600s;\n\
    proxy_connect_timeout   3600s;\n\
    proxy_buffering         off;\n\
    client_max_body_size    10M;\n\
    proxy_pass              http://127.0.0.1:CWS;\n\
  }\n\n\
  location SUBFOLDERapi {\n\
    proxy_pass              http://127.0.0.1:CWS;\n\
    proxy_set_header        Host \$host;\n\
    proxy_set_header        X-Real-IP \$remote_addr;\n\
    proxy_set_header        X-Forwarded-For \$proxy_add_x_forwarded_for;\n\
    proxy_set_header        X-Forwarded-Proto \$scheme;\n\
    client_max_body_size    65M;\n\
  }\n\n\
  location SUBFOLDERwebsocket {|g" \
        /defaults/default.conf

# Interposer + fake-udev rebuilt against the comprehensive-fixes protocol.
COPY --from=native-builder /tmp/selkies/addons/js-interposer/selkies_joystick_interposer.so \
        /usr/lib/selkies_joystick_interposer.so
COPY --from=native-builder /tmp/selkies/addons/fake-udev/libudev.so.1.0.0-fake \
        /opt/lib/libudev.so.1.0.0-fake

# comprehensive-fixes' unified server fronts the websocket with basic auth by
# default (ubuntu/mypasswd unless SELKIES_BASIC_AUTH_* is set). Browsers
# cannot attach an Authorization header to a WebSocket upgrade; the server
# now skips basic auth on websocket paths once a master token is provisioned,
# but a deployment without token provisioning still locks browser clients
# out unless this is disabled. (The LSIO images never set the enable env, and
# nginx-level auth via CUSTOM_USER/PASSWORD is available instead.)
ENV SELKIES_ENABLE_BASIC_AUTH=false
```

</details>

rmounce:
Two more findings from testing on the lsio-based image, both about failure opacity rather than the happy path:

**1. WebSocket upgrades rejected by basic auth are silent on both ends.** In `_auth_middleware`, a request with no `Authorization` header gets a 401 with no server-side log (unlike the disallowed-Origin and invalid-credentials branches, which both log warnings). For a WebSocket upgrade this is doubly invisible: browsers can't attach an `Authorization` header to a WS handshake, and the page only sees a generic connection failure (Firefox reports `NS_ERROR_WEBSOCKET_CONNECTION_REFUSED`; the status code isn't exposed to JS). Since `enable_basic_auth` defaults to true (ubuntu/mypasswd), a deployment that doesn't explicitly set `SELKIES_ENABLE_BASIC_AUTH=false` has a data websocket no browser can reach, and neither side logs why — this cost us a debugging session. Suggestions: log rejected WS upgrades at warning level like the Origin check does, and consider whether the Basic gate should apply to websocket paths at all when token auth is in play (the token is already validated in `data_ws_handler`).

**2. When pixelflux fails to import, capture start dies with an unrelated-looking `TypeError` instead of the clean degrade.** The import fallback sets `ScreenCapture = CaptureSettings = None` and logs a good warning at startup — but in `_start_capture_for_display_impl`, `_get_capture_settings()` (which instantiates `CaptureSettings()`) runs before the `X11_CAPTURE_AVAILABLE` guard, so "take control" crashes with `'NoneType' object is not callable`, long after the informative startup warning has scrolled away. Hoisting the `X11_CAPTURE_AVAILABLE` check to the top of `_start_capture_for_display_impl` would surface the real cause at the point of failure. (This is how we experienced the libwayland issue reported on linuxserver/pixelflux#18 — the crash gave no hint the import had failed.)

ehfd:
Please continue the conversation in https://github.com/selkies-project/selkies/issues/227.

PR #255, opened 2026-06-14
fix(clipboard): send current clipboard to freshly connected clients

DL6ER:

No tracking issue.

The outbound clipboard monitor (`WebRTCInput.start_clipboard`) only sends when the clipboard *changes*. A client that connects - or reconnects after a page reload or a resumed session - while the clipboard is unchanged since the previous client left never receives the existing clipboard contents until the next change. The server -> client clipboard bridge then appears dead even though the server holds valid clipboard text.

This pull request pushes the current clipboard to a client the moment it connects, the same way the current cursor is already pushed on connect.

- Add `DataStreamingServer.send_current_clipboard(websocket, raddr)`, mirroring the existing `send_current_cursor`. It reads the current clipboard via the input handler and sends it once, directly to the freshly connected socket.
- Call it right after `send_current_cursor` in the data WebSocket handler.
- Outbound `text/plain` only and within `CLIPBOARD_CHUNK_SIZE`; binary and large payloads keep flowing through the monitor/multipart path. It honors the `enable_clipboard` direction (`true`/`out`) and is a no-op when outbound clipboard is disabled. Exceptions are caught and logged, never raised into the connect path.
- No new dependencies; reuses `read_clipboard` and `CLIPBOARD_CHUNK_SIZE`.

Tested on an EL/XFCE (Python 3.9) webtop with an end-to-end clipboard test that parks and wakes the session (client reconnect while the server clipboard is unchanged): the reconnected client now receives the existing clipboard, which it did not before this change. No change to behavior for already-connected clients (the monitor path is untouched).

- Resetting the monitor's change baseline so it re-broadcasts on the next tick: only fires on the next tick/change and is subject to broadcast backpressure skips; a direct per-socket send on connect is deterministic and independent of tick timing.
- A bounded periodic re-send for a window after connect: turned out unnecessary - a single connect-time send was sufficient in testing.

**Additional context**

Only outbound `text/plain` clipboard is handled here, consistent with the connect-time cursor send.

### DIFF
--- a/src/selkies/selkies.py
+++ b/src/selkies/selkies.py
@@ -1115,6 +1115,56 @@ class DataStreamingServer(BaseStreamingService):
         except Exception as e:
             data_logger.warning(f"Failed to send current cursor to client {raddr}: {e}")
 
+    async def send_current_clipboard(self, websocket, raddr):
+        """Re-assert the current server clipboard to a freshly connected client.
+
+        The clipboard monitor (WebRTCInput.start_clipboard) only emits on
+        *change*, so a client that connects - or reconnects after a page
+        reload or a resumed session - when the clipboard is unchanged since the
+        previous client left never receives the existing contents, and the
+        server->client clipboard bridge looks dead. A single connect-time send
+        is not reliable either: the client may only start reading a moment
+        later, or briefly re-assert its own local clipboard right after load and
+        overwrite the value before it reads it back. So re-send the current text
+        clipboard directly to this socket periodically for a bounded window
+        after connect, until the socket closes or leaves the client set.
+        Outbound text only; binary and large payloads stay on the
+        monitor/multipart path. Runs as a background task (see ws_handler), so
+        it never blocks connection setup.
+        """
+        ih = self.input_handler
+        if not ih or getattr(ih, "enable_clipboard", "") not in ("true", "out"):
+            return
+        from .input_handler import CLIPBOARD_CHUNK_SIZE
+        # ~60s window, re-asserting every 2s: long enough to overlap a slow
+        # client's initial read loop, short enough not to linger.
+        deadline = time.monotonic() + 60.0
+        first = True
+        while time.monotonic() < deadline:
+            if not first:
+                await asyncio.sleep(2.0)
+            first = False
+            if websocket not in self.clients:
+                return
+            try:
+                clip_data, clip_mime = await ih.read_clipboard(use_binary=False)
+                if not clip_data or clip_mime != "text/plain":
+                    continue
+                clip_bytes = clip_data.encode("utf-8") if isinstance(clip_data, str) else clip_data
+                if not (0 < len(clip_bytes) < CLIPBOARD_CHUNK_SIZE):
+                    continue
+                encoded = base64.b64encode(clip_bytes).decode("ascii")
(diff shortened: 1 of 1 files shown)


DL6ER:
Thanks for the review. Addressed:

- Spawn the clipboard send via `asyncio.create_task` so the connection setup is not blocked by `read_clipboard` (which shells out to `xclip`/`wl-paste`). `server_settings` and stream startup are no longer delayed.
- Demoted the per-connect log line to `debug` to avoid noise on frequent reconnects.

Kept intentionally, for consistency with the existing `send_ws_clipboard_data`: the `< CLIPBOARD_CHUNK_SIZE` single-frame cutoff (payloads `==` the chunk size go multipart there too) and the function-local `CLIPBOARD_CHUNK_SIZE` import (avoids the `input_handler` circular import). Did not add an empty-clipboard "clear" send, since a freshly connected client has no prior clipboard to clear.

DL6ER:
Heads-up for reviewers: the failing `py-build` / `gst-web` *image build & publish* checks are not caused by this change. They fail identically on `main` (the image-build workflow currently can't resolve its base `ghcr.io/selkies-project/selkies/selkies-web:main`), and this PR only touches `src/selkies/*.py` - none of the image Dockerfiles. The code check (`validate`) passes.

DL6ER:
Revised: replaced the single connect-time send with a **bounded periodic re-send** (background task, ~60s window, every 2s, stopping as soon as the socket closes or leaves the client set).

A one-shot connect-time send turned out not to be reliable in practice. In a controlled measurement on our park-and-restore end-to-end test (`managedsave`/restore + client reconnect with the server clipboard unchanged), the one-shot version failed the server->client clipboard assertion **2 of 2** runs (the client either hadn't started reading yet, or briefly re-asserted its own local clipboard right after load and overwrote the value before reading it back), while the bounded periodic re-send passed **4 of 4** runs.

Still outbound `text/plain` only (binary/large stays on the monitor/multipart path) and still spawned as a background task, so it never blocks connection setup.

PMohanJ:
Hey @DL6ER. Thanks for the PR. Highly appreciate it.
I'd to point you to this UI code: https://github.com/selkies-project/selkies/blob/main/addons/selkies-web-core/selkies-ws-core.js#L2925-L2926 which logically does what this PR is trying to achieve. Since you mentioned webtop images which support websockets streaming mode(as of now), have you tested this flow - client requesting for clipboard read up on ws connection.

Please let us know. I think this flow `client requesting sever` is much simpler than server trying to send clipboard data.

DL6ER:
You are right, and thanks for the pointer. I traced it:

- On `main`, `selkies-ws-core.js` sends `cr` on every ws (re)connect (the lines you linked), and the server handler reads and pushes the current clipboard unconditionally in response. That covers the freshly-connected / reconnected client cleanly, so a server-initiated push is redundant on `main`.
- The reconnect-with-stale-clipboard symptom I hit was on an older pinned webtop generation that predates the backend rewrite (#236): that tree ships no `selkies-ws-core.js`, so there is no client `cr`-on-connect and nothing re-reads the clipboard after reconnect. The fix there is to move that deployment onto the current core, not to add a server push.

So this PR is solving a problem that no longer exists on `main`. The client-requests-server flow is the simpler and correct design - closing this. Thanks for the review.

PR #267, opened 2026-07-06
perf(input): reduce per-mouse-move overhead (X11 flush, client rAF coalescing)

rmounce:
## Problem

Two independent sources of per-mouse-move overhead on the input path:

1. **Server**: `send_x11_mouse()` and `send_mouse()` call `self.xdisplay.sync()`
   after injecting motion, which blocks on a full X server round-trip per
   event. On the absolute (non-relative) path this sync runs even though
   pynput drives the motion on its own display connection, so it round-trips
   a connection with nothing pending. This blocking happens on the event loop
   the video sender shares.
2. **Client**: every `mousemove` DOM event produced a websocket send. At high
   polling rates (gaming mice easily exceed 500 events/s) this floods the
   socket with intermediate positions the server will immediately overwrite.

## Fix

1. Use `flush()` instead of `sync()` — it delivers queued input in order
   without waiting for a reply, removing the per-move event-loop stall.
   (`comprehensive-fixes` makes the same change; this extracts it so it can
   land independently.)
2. Coalesce mouse-move sends to one per animation frame: intermediate
   positions within a frame are collapsed to the latest, non-move input is
   unaffected, and the final position is always flushed so no endpoint is
   lost.

## Testing

Running in our production deployment (lsio-based images) for two weeks;
measurably smoother pointer under load with no observed input loss or
reordering. Both hunks apply to `main` unchanged (cherry-picks; the
surrounding code is identical between the branches). Not yet runtime-tested
against `main` itself — a main-based test build on production NVIDIA hardware
is in progress, results to follow on this PR.

### DIFF
--- a/addons/selkies-web-core/lib/input.js
+++ b/addons/selkies-web-core/lib/input.js
@@ -1727,8 +1727,58 @@ export class Input {
                 this.buttonMask &= ~mask;
             }
         }
-        var toks = [ mtype, this.x, this.y, this.buttonMask, 0 ];
-        this.send(toks.join(","));
+        if (event.type === 'mousemove' || event.type === 'pointermove') {
+            // Coalesce high-frequency motion. A 1000 Hz mouse would otherwise emit
+            // ~1000 tiny WS messages/sec, congesting the uplink (felt as input lag
+            // over the internet) and starving the server's single-threaded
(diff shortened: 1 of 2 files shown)


ehfd:
Incorporated into https://github.com/selkies-project/selkies/pull/254.

PR #268, opened 2026-07-07
fix(input): pointer pinned at (0,0) when frames present on the video sink in manual-resolution mode

rmounce:
In manual-resolution mode `_applySinkCoordinates` maps pointer coordinates against `#videoCanvas`. ws-core hides that canvas (`display: none`) whenever decoded frames are presented on the `<video>`/worker sink (`presentFrameToVideo` / `activateWorkerSinkDisplay`), so the canvas measures 0×0 the moment any frame arrives. The function then set `x = y = 0` and still returned `true`, suppressing the windowMath fallback — every subsequent pointer event streamed `m,0,0` and the remote pointer was pinned at the top-left corner.

On a mostly static desktop this masquerades as "right-click breaks the mouse": decoder warm-up frames paint on the (visible) canvas, nothing changes on screen, input works — then the right-click menu appearing produces the first frame on the video sink, the canvas is hidden, and the pointer dies. A window resize heals it (resize handlers re-show the canvas) until the next frame. With continuously changing content the pointer is dead immediately.

Since the presenter mirrors the canvas box verbatim onto the visible sink, the fix measures the visible mirror (`#videoStream` / `#videoWorkerCanvas`) when the canvas rect is zero (the canvas buffer dimensions stay authoritative for scaling), falls back to the last valid measurement, and only if neither exists returns `false` so the caller uses windowMath instead of claiming success with (0, 0).

Tested on the comprehensive-fixes stack (selkies 40a5312 + pixelflux 968a4f4 + pcmflux 6aa227e) on an lsio-based image with a Firefox client at manual resolution 1920×1080: reproduced the pinned pointer, verified the fix gives correct absolute coordinates through right-click menus and continuous screen updates, with no offset on a fresh session.


ehfd:

PR #269, opened 2026-07-07
fix(websockets): don't reject same-origin WS upgrades behind port-stripping proxies

rmounce:
`_is_ws_origin_allowed` compares the `Origin` header against the forwarded `Host`. Behind a reverse proxy configured with `proxy_set_header Host $host` (nginx `$host` strips the port — this is what linuxserver.io's bundled config does), the browser's `Origin` carries an explicit port (e.g. `https://example.com:3001`) while the forwarded `Host` has none, so the exact string comparison fails and every same-origin websocket upgrade is rejected — with no hint client-side, since the 403 happens before the upgrade.

Fix: when the forwarded `Host` carries no port, fall back to comparing hostnames only. Exact host:port matching is preserved whenever the proxy forwards a port. Checked against IPv6 literals, `Origin: null`, and genuine cross-site origins (still rejected).

Found and tested while second-eye testing #254 on an lsio-based image (their nginx proxies the websocket with `Host $host` on port 3001).


ehfd:

PR #273, opened 2026-07-08
fix(video): sync the video sink on resize instead of waiting for a frame

rmounce:
On a window resize, the canvas-style writers (`applyManualCanvasStyle` / `resetCanvasStyle`) unconditionally re-show the main canvas and rewrite its box; the active video sink (`<video>`/worker canvas) is only re-hidden behind it and re-mirrored to the new geometry by the present paths — i.e. on the **next decoded frame**. On a static remote there is no next frame, so after a resize the stale canvas content (typically the connection-time picture) covers the live sink, correctly scaled — a "freeze-frame" that tracks the window — until motion produces a frame. Reproduced in both Chromium (MSTG sink) and Firefox (worker canvas sink).

Fix: when the active sink has proven it renders, sync it directly from the style writers — mirror the fresh canvas box onto it and re-hide the canvas immediately, no frame required. Warm-up behavior (canvas visible until the sink's first render) is unchanged.

Tested on the LSIO-based image against comprehensive-fixes (`da1b55e`): resize with a static remote now stays on the live stream in both browsers.


ehfd:
Incorporated into #254

PR #277, opened 2026-07-14
fix(display): honor manual resolution on Wayland and pre-client capture paths

rmounce:
Three related gaps around a configured manual resolution (`SELKIES_MANUAL_WIDTH/HEIGHT`), all masked on X11 by the display server being sized externally:

- `SelkiesStreamingApp` fell back to a hardcoded 1024x768 for capture paths that run before any client sizes the display (viewer-driven capture on a controller-less server). On Wayland the fallback is not cosmetic: the capture start actively resizes the compositor output down to it. The fallback now seeds from the manual resolution.
- The WebRTC-mode startup applied the manual resolution via `resize_display()` unconditionally, dialing RandR/xrandr against a nonexistent X server on Wayland (`Native RandR resize for '1920x1080' failed ... Cannot resize display via xrandr`). Now gated on the backend.
- The WebRTC-mode `MediaPipeline` kept its constructor-default 1920x1080 regardless of the manual resolution, so on X11 a non-1920x1080 manual resolution captured the wrong geometry until the first client resize. Its dimensions now seed from the manual resolution.

Validated on the LSIO-based test image: with `PIXELFLUX_WAYLAND=true`, `SELKIES_MANUAL_WIDTH/HEIGHT=1920/1080` and no controller connected, the viewer-driven capture now comes up at 1920x1080 (previously 1024x768).


rmounce:
A note on the gemini review above, since it looks alarming out of context: it was generated against a stale merge base. `comprehensive-fixes` was force-pushed (21e0cfe → d2149bc) shortly after this PR was opened, so the review covered the base branch's own new content (chunked uploads, the input.js gamepad changes, the settings sanitizer refactor) rather than this PR's 27-line diff. All six inline comments anchor to files or lines this PR doesn't modify.

For completeness we checked the one "critical" claim (startup `TypeError` in settings.py when parsing manual dimensions) against the current tip: it's a false positive — int-typed settings are stored as plain ints (the `(value, locked)` tuple treatment applies to bool/range settings), and the settings pipeline loads cleanly with `SELKIES_MANUAL_WIDTH/HEIGHT` set.

The branch has since been rebased onto d2149bc, so the diff shown is now the intended three fixes only.

ehfd:
Applied to #254 with improvements.

PR #278, opened 2026-07-14
feat(dashboard): make the sidebar toggle handle draggable with persisted position

junkerderprovinz:

No existing issue; small self-contained dashboard UX feature.

The sidebar toggle handle (`.toggle-handle` in the classic dashboard) is hard-fixed at 50% of the viewport height on the left edge. Whatever application UI happens to sit under that spot is permanently covered by the handle and its hover/click zone, and the user has no way to move it out of the way. The KasmVNC-style control bar handle that many users come from was draggable and remembered its position, so this is also a parity gap for people migrating to Selkies.

This PR makes the handle vertically draggable along the left edge and persists the chosen position across reloads. A plain click keeps toggling the sidebar exactly as before.

All changes are confined to `addons/selkies-dashboard/src/components/Sidebar.jsx` (+81/-1); no CSS, protocol, or server changes.

- The drag implementation reuses the exact pointer-capture pattern already in the codebase for the virtual keyboard button (`dragInfo` / `handlePointerDown/Move/Up` in `Sidebar.jsx`) and `PlayerGamepadButton.jsx`: `setPointerCapture` on pointerdown, the existing `DRAG_THRESHOLD` (10 px) before entering drag mode, and the `hasDragged` flag that suppresses the click event fired right after a real drag. Pointer events mean mouse and touch both work; `touchAction: 'none'` (inline, same as the keyboard button) keeps touch drags from scrolling.
- Only the pointer's Y is applied; the handle stays on the left edge. The position is stored as a **percentage of the viewport height** under the app-prefixed localStorage key `sidebarToggleTopPct` (via the existing `getPrefixedKey` helper), so window resizes keep the position proportional. It is written once on pointerup, restored on mount, and clamped on both restore and drag so the 60 px handle always stays fully inside the viewport (the viewport may have shrunk since the value was saved). The default remains the current 50%.
- The position is applied as an inline ``style={{ top: `${pct}%` }}``; `Overlay.css` is untouched, so `left: 0`, `transform: translateY(-50%)`, the iOS `@supports` width bump, and all indicator styling keep working unchanged. The `isToggleVisible` gating (hidden for viewer-role clients) is unaffected.

Verification: built `addons/selkies-web-core` (`npm install && npm run build`), copied `dist/selkies-core.js` into the dashboard `src/` as the Dockerfile does, then built `addons/selkies-dashboard` (`npm install && npm run build`) — vite build completes cleanly. `npx eslint src/components/Sidebar.jsx` reports 0 errors (the 5 pre-existing `react-hooks/exhaustive-deps` warnings are untouched). The drag/click-suppression logic is a line-for-line mirror of the in-repo `PlayerGamepadButton`/keyboard-button pattern that is already shipping. I have not yet exercised this in a live Selkies session, so a quick interactive smoke test (drag the handle, click it, reload, resize) by a reviewer with a running stack would be appreciated.

- Persisting pixels instead of a percentage: breaks on window resize and on different screen sizes; percentage keeps the position proportional for free with no resize listener.
- Making the handle draggable to the right edge as well (full KasmVNC parity): deliberately out of scope to keep the diff minimal; the sidebar itself only opens from the left today.
- A settings entry for the position: dragging the thing you want to move is more direct and adds no settings surface.

**Additional context**

Storing the value once on pointerup (rather than every pointermove) keeps localStorage writes to one per drag. A click without crossing the 10 px threshold behaves byte-for-byte like the current handler.

### DIFF
--- a/addons/selkies-dashboard/src/components/Sidebar.jsx
+++ b/addons/selkies-dashboard/src/components/Sidebar.jsx
@@ -670,8 +670,14 @@ function useConditionalSetting(spec, serverSettings, ctx, deps) {
 // inside the viewport, expressed as a percentage of the viewport height.
 const TOGGLE_HANDLE_HEIGHT_PX = 60; // keep in sync with .toggle-handle in Overlay.css
 const clampToggleHandleTopPct = (pct) => {
-  const halfHandlePct = (TOGGLE_HANDLE_HEIGHT_PX / 2 / window.innerHeight) * 100;
-  return Math.min(100 - halfHandlePct, Math.max(halfHandlePct, pct));
+  const safePct = Number.isFinite(pct) ? pct : 50;
+  // A zero/undefined viewport height (headless, pre-layout, test env) would make
+  // halfHandlePct Infinity and clamp everything to a broken edge, so fall back
+  // to a plain 0..100 clamp until a real height is known.
+  const vh = window.innerHeight;
+  if (!vh || !Number.isFinite(vh)) return Math.min(100, Math.max(0, safePct));
+  const halfHandlePct = (TOGGLE_HANDLE_HEIGHT_PX / 2 / vh) * 100;
+  return Math.min(100 - halfHandlePct, Math.max(halfHandlePct, safePct));
 };
 
 function Sidebar() {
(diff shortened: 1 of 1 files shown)


ehfd:
Open to a merge, but I do need the issues solved or disproved.

ehfd:
Incorporated into and superseded by https://github.com/selkies-project/selkies/commit/c1ae1417d12e01337bddb8810ab1d7b252b6a2c2.

junkerderprovinz:
Thanks @ehfd — glad the draggable toggle was useful and made it into c1ae1417 (with the co-author credit, appreciated). Closing this since it's incorporated upstream now. Onward to v2.0.0!

Issue #263, opened 2026-07-02
Implement different WebRTC protocol topologies such as ICE-TCP/UDP


ehfd:
Can be AI-driven: Yes, as long as all cases are tested carefully in real-life by people as well as automated tests.
Expected to be workable with: Fable 5, GPT-5.6 Sol, Kimi-K3, Opus 5, or equivalent (risky with Opus 4.8 or under).
Prompt or reference: Comments below, https://neko.m1k1o.net/docs/v3/configuration/webrtc, https://github.com/pion/ice, https://github.com/pion/webrtc.

---

Different protocol topologies to TURN and STUN (such as ICE-TCP/UDP)

https://neko.m1k1o.net/docs/v3/configuration/webrtc

Pion provides various WebRTC configurations and protocols including EPR, UDPMUX, TCPMUX, NAT1TO1, ICE-LITE, ICE-TCP, etc. These techniques allow more setup flexibility in addition to TURN/STUN and allow limiting port ranges or using a single port for many numbers of connections.

Same as TCPMUX and UDPMUX in Pion and Neko - more information in https://neko.m1k1o.net/docs/v3/configuration/webrtc

junkerderprovinz:
Hi @PMohanJ, now that the ICE stack lives in-tree (e171691), the smallest slice of this issue, a configurable UDP port range for host candidates, has become a fairly contained change: a bind-in-range loop where the host sockets currently take ephemeral ports, plus one setting threaded through the same way `webrtc_public_ip` was. Since I worked on the NAT1TO1 part of that commit I would be happy to pick up that slice as a standalone PR, if you are not already on it and @ehfd agrees. The bigger asks here (ICE-TCP, mux, lite) I would leave untouched. Just say the word either way.

PMohanJ:
Sorry for the delayed response.
Sure if you wish to pick up that part, feel free to draft a PR, later I'll rebase the rest of the parts on that branch. It'll take few more day for me get back on track.

junkerderprovinz:
Quick update after digging into the host-candidate binding: the ephemeral ports are allocated inside aioice's `gather_candidates` (`Connection` has no port-range knob in 0.10.x or on `main`), so the port-range slice cannot land as a selkies-only change the way the public-IP rewrite did. It needs the range in aioice first.

[...]

PR #280, opened 2026-07-16
feat(webrtc): advertise a configured public IP in host ICE candidates (NAT1TO1)

junkerderprovinz:
### What

Adds an optional `webrtc_public_ip` setting (`--webrtc-public-ip` / `SELKIES_WEBRTC_PUBLIC_IP`). When set, Selkies advertises that IPv4 in its **host** ICE candidates — the [Pion `NAT1TO1`](https://github.com/pion/webrtc/blob/master/settingengine.go) idea — so a host behind **static 1:1 NAT** (e.g. a cloud instance whose private address maps one-to-one to a fixed public/elastic IP, WebRTC UDP ports forwarded) is reachable directly instead of only via a TURN relay. Empty (default) = unchanged behaviour.

Partially addresses #263. See the scope note below for what this deliberately does **not** try to do.

### How

`RTCApp.munge_sdp` is the single rewrite point: the offer is non-trickle (`setLocalDescription(createOffer())` embeds all gathered candidates), so every host candidate passes through here. The new `_rewrite_host_candidate_ip` replaces the connection-address of IPv4 `typ host` candidates with the configured IP and leaves **srflx (STUN), relay (TURN), their `raddr`, and all ports / foundations / priorities untouched** — so hole-punching and TURN fallback keep working if the direct path is blocked. It is a no-op when the value is empty or not a valid dotted-quad IPv4.

### Scope (honest)

The other topologies in #263 — ephemeral UDP **port range**, **ICE-TCP**, **UDP/TCP mux** (single port), **ICE-LITE** — are **not achievable in Selkies alone**. aiortc is vendored here, but ICE goes through the external `aioice` dependency, which gathers UDP host candidates on ephemeral ports and exposes no port / IP / TCP / lite knobs (checked against aioice 0.10.1 and `main`). Those would each need an upstream `aioice` contribution, not a Selkies change. NAT1TO1 is the one piece Selkies can own end to end. (Also worth noting: the single-port reverse-proxy story already has an answer today via `turns:443` TCP/TLS.)

### Tests / verification

- **Unit** (`src/selkies/tests/test_nat1to1_sdp.py`): pure-string tests for the rewrite — host IPv4 rewritten; srflx/relay/raddr/IPv6/ports/foundations preserved; empty/invalid IP no-op; idempotent. No browser/ICE/network, so it runs as a fast gate.
- **Live**: run a container with `SELKIES_WEBRTC_PUBLIC_IP=<public-ip>` on a 1:1-NAT host; in `chrome://webrtc-internals` the local **host** candidate shows the public IP and the selected candidate pair reaches **succeeded** on it, vs. the private IP + TURN fallback when unset.

Happy to adjust naming or split the docs out if you'd prefer.


ehfd:
My thought about #263 is that we will eventually integrate aioice (just like we did for aiortc) into the codebase and make our own edits. I would personally prefer everything to be done at once.

junkerderprovinz:
Thanks @ehfd. I agree the aioice integration (mirroring the aiortc approach) is the right long-term home for the full set of topologies in #263 (NAT1TO1, TCP/UDP MUX, ICE-TCP/LITE, etc.).

I kept this PR deliberately narrow: it's a self-contained SDP host-candidate rewrite (~29 lines in `rtc.py` plus one setting and a test) that only advertises a configured public IPv4 in host candidates. It doesn't touch the ICE machinery, so it gives users static 1:1 NAT support today without waiting on the larger refactor, and it can be dropped or superseded cleanly once aioice is vendored in and handles NAT1TO1 natively.

That said, it's your call — happy to leave it in as an interim step, or to close it and fold the behaviour into the aioice work when that lands if you'd rather do everything at once. Just let me know which you prefer.

I've also addressed the automated review notes in the latest push:
- aligned the flag name in `docs/firewall.md` with the actual setting (`--webrtc-public-ip`), and
- moved the IP validation to the stdlib `ipaddress` module, in both the implementation and the test.

One note on CI: the red checks look unrelated to this change — the gst-web image build fails with `unable to prepare context: path "addons/gst-web" not found`, and this PR only touches Python and docs.

ehfd:
https://github.com/selkies-project/selkies/commit/e171691464bd313d01cb1e505d409838fca72786

Implemented.

junkerderprovinz:
That's great, thank you! I can confirm e171691 covers everything this PR set out to do (the `webrtc_public_ip` setting plus the per-family host-candidate rewrite, with STUN/TURN candidates untouched), and folding it into the vendored ICE work is exactly the "all at once" route you described. Appreciate the co-author credit. Closing this as superseded.

PR #281, opened 2026-07-18
docs(faq): document missing video output while the screen is locked

junkerderprovinz:
**Reference the issue numbers and reviewers**

#174 @ehfd

**Explain relevant issues and how this pull request solves them**

This pull request documents the known limitation discussed in #174: when the screen is locked on an existing desktop, Selkies produces no video output. As established in that thread, the display manager's lock switches to a greeter on a separate display or virtual terminal (observed with LightDM spawning a second X server on `:1`), so Selkies capturing the original `DISPLAY` loses output until the screen is unlocked. Since documentation was explicitly requested in that issue, this adds a FAQ entry covering the cause and the workarounds. Documents #174; feel free to close that issue if this satisfies the documentation requirement.

**Describe the changes in code and its dependencies and justify that they work as intended after testing**

Adds one new entry to `docs/faq.md`, matching the existing entry pattern (`##` heading, `<details markdown>` block with an `Open Answer` summary). It explains why the video output stops on screen lock and lists the workarounds in order: disable screen locking and any autostarted locker (`light-locker`, `xscreensaver`, `gnome-screensaver`), run `xset s off -dpms`, add `Option "HardDPMS" "False"` in `/etc/X11/xorg.conf` for NVIDIA DPMS blanking, or use a container/session without a display manager. Verified that the Markdown renders correctly with the `md_in_html` extension used for the existing entries: all eight `<details>`/`</details>` pairs and summaries are balanced. Documentation-only change; no code or dependencies are affected.

**Describe alternatives you've considered**

Handling the lock screen transition in code was ruled out in #174 ("There is not much that can be reliably done here"), since the greeter runs on a different display that Selkies cannot follow, so documenting the limitation and workarounds is the agreed approach.

**Additional context**

The new entry is placed directly after the existing display-manager-related entry ("The web interface refuses to start up in the terminal after rebooting...") to keep session/display-manager topics grouped together.

 - [x] I confirm that this pull request is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please file the pull request there.
 - [x] I confirm that this pull request has been tested thoroughly and to the best of my knowledge that additional unintended problems do not arise.
 - [x] I confirm that the style of the changed code conforms to the overall style of the project.
 - [x] I confirm that I have read other open and closed pull requests and that duplicates do not exist.
 - [x] I confirm that I have justified the need for this pull request and that the changes reflect the fix for the specified problem.
[...]

Issue #174, opened 2025-01-24
When the screen is locked, Selkies will have no video output


Resulte:
**Is your enhancement request related to a problem? Please describe.**
When the screen is locked, Selkies will have no video output:

![Image](https://github.com/user-attachments/assets/31b15918-95f0-4be6-afa6-9fee8b29f1ad)
At this time, I am unable to perform any operations through Selkies.

**Describe the solution you'd like**
I want to be able to use Selkies even when the screen is locked, including unlocking the screen by Selkies.
Have any ideas to solve this problem?

 - [x] I confirm that this issue is relevant to the scope of this project. If you know that upstream projects are the cause of this problem, please raise the issue there.
 - [x] I confirm that I have read other open and closed issues and that duplicates do not exist.
 - [x] I confirm that I have described the solution as well as the problem in detail, and if possible, explained how to solve the problem.
 - [x] I confirm that no portion of this issue contains credentials or other private information, and it is my own responsibility to protect my privacy.
 - [x] I confirm that the authors of this issue does not willfully breach or infringe legal regulations, in any and all global law, regarding trademarks, trade names, logos, patents, or any and all other forms of external intellectual property, as well as adhering to software license terms of open-source and proprietary software projects.

PMohanJ:
Based on what you explained, I assume that you're basically accessing an existing desktop with Selkies.
Screen lock and unlock falls under OS level configurations; I highly doubt we'd support such, because Selkies is more focused on running desktops in containerised environments. @ehfd correct me here.

Nonetheless, it's nice feature to have. I guess Sunshine support this.

ehfd:
I have no information on the desktop environment and OS, etc. First need those things. Otherwise, I don't see a good reason why it can't work with X11.

Resulte:
> I have no information on the desktop environment and OS, etc. First need those things. Otherwise, I don't see a good reason why it can't work with X11.我没有关于桌面环境和操作系统等信息。首先需要这些。否则，我看不出为什么它不能与 X11 兼容。

OS version: ubuntu 22.04
Desktop environment: x11 + gnome.
In my test, Selkies can run normally in this environment, but when I lock the screen, Selkies will not have any video output and cannot perform any operations.

ehfd:
Thank you for the information.

First, make sure that you are running `selkies-gstreamer` as root. This is required in order to communicate with the session manager.

Then, I suggest (less likely to be the culprit):

```
Option "HardDPMS" "False"
```

[...]

PR #283, opened 2026-07-28
feat(example): opt-in flag to keep capture alive when the screen is locked

junkerderprovinz:
## Summary

Addresses #174.

When Selkies captures the X framebuffer and a screensaver, DPMS blanking, or a desktop screen locker blanks the display, the WebRTC stream goes black and the user cannot interact with the session to unlock it. This adds an opt-in flag to the reference example container so operators can keep capture alive across those events.

## What this changes

A new environment variable, `SELKIES_DISABLE_SCREEN_BLANKING`, wired into `addons/example/entrypoint.sh`:

- Defaults to `false`. When unset or false, behavior is exactly as before, so nothing changes for existing users.
- When set to `true` (or `1`), right after the X server is ready and before the Xfce4 session starts, it:
  - runs `xset s off`, `xset s noblank`, and `xset -dpms` to turn off the built-in X screen saver, screen blanking, and DPMS power management;
  - writes `Hidden=true` XDG autostart overrides for the common lockers (`light-locker`, `xscreensaver`, `xfce4-screensaver`, `gnome-screensaver`, `mate-screensaver`, `xss-lock`) so the desktop session does not autostart them and blank the framebuffer on lock.

The variable is also surfaced in `docker-compose.yml` alongside the other `SELKIES_*` settings for discoverability. Naming and the on/off check follow the existing convention in the same entrypoint (for example `SELKIES_NGINX_OVERRIDE` and the lowercased `true` comparisons used for `SELKIES_ENABLE_HTTPS`).

## Notes

- Opt-in only. Default behavior is unchanged.
- This complements the documentation in PR #281, which describes the same workaround for operators who prefer to configure it by hand.
- `xset` is already available in the example image via `x11-xserver-utils`, and `xfce4-goodies` can pull in `xfce4-screensaver`, so the locker masking is relevant to this image.

## Testing

- `bash -n addons/example/entrypoint.sh` passes.
- With the flag unset, the entrypoint is equivalent in behavior to before, since the guarded block is skipped.

### DIFF
--- a/addons/example/entrypoint.sh
+++ b/addons/example/entrypoint.sh
@@ -51,6 +51,21 @@ echo 'Waiting for X Socket' && until [ -S "/tmp/.X11-unix/X${DISPLAY#*:}" ]; do 
 # Preset the resolution
 selkies-resize 1920x1080
 
+# Opt-in: disable screen blanking, DPMS, and screen lockers so a locked/blanked
+# session does not stop framebuffer capture (default off, current behavior unchanged).
+# See https://github.com/selkies-project/selkies/issues/174
+if [ "${SELKIES_DISABLE_SCREEN_BLANKING:-false}" = "1" ] || [ "$(echo "${SELKIES_DISABLE_SCREEN_BLANKING:-false}" | tr '[:upper:]' '[:lower:]')" = "true" ]; then
+  # Turn off the built-in X screen saver, screen blanking, and DPMS power management
+  xset s off || true
+  xset s noblank || true
+  xset -dpms || true
+  # Prevent common desktop screen lockers from autostarting and blanking the framebuffer on lock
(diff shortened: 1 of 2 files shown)


ehfd:
Nice! I'll review it when I have time.

ehfd:
Unfortunately, I am aware that the example image does not particularly need this because all screen lock features are disabled. I will take in documentation inputs, though.

junkerderprovinz:
Thanks @ehfd, that makes sense. With the screen-lock features disabled in the example image, the opt-in flag isn't needed there, so I'll close this in favor of the documentation route. The behaviour and the workarounds are covered in #281 (now rebased and ready to merge). Happy to revisit if a concrete use-case for the flag comes up outside the example image.

PR #296, opened 2026-08-09
fix(input): send the character keysym directly for typed and pasted uppercase

### LINKED ISSUE #295: Clipboard re-type path loses uppercase when the X keymap does not bind Shift
### Summary

The clipboard re-type / soft-input path capitalises by sending `Shift_L` down, then the LOWERCASE keysym, then `Shift_L` up, relying on the X server's Shift modifier. On any deployment whose X keymap does not bind Shift as a real modifier, that Shift never engages and every uppercase letter arrives lowercase, while the physical-keyboard path is fine because it sends the uppercase keysym directly (for example `XK_A`) and lets the server resolve the shift level.

Concretely: on the LinuxServer Selkies base, Xvfb starts with no keymap. When Firefox has no secure-context async clipboard and falls back to re-typing the paste through this path, `Big Chicken A Fast Food Conspiracy` pastes into the session as `big chicken a fast food conspiracy`. Lowercase letters, digits and spaces are untouched, only capitals (and shifted symbols) drop out. Same family as #257 and #214. #254 helped Firefox by restoring the byte-exact selection path, but the re-type fallback still has this weakness.

### Suggested fix

Make the re-type path send the uppercase keysym directly, the way the physical-keyboard path already does, instead of `Shift_L` plus the lowercase keysym. Then capitalisation no longer depends on a Shift modifier that a minimal or incomplete keymap may not bind, and the paste keeps its case regardless of the deployment keymap.

I understand the X11 keymap itself is deployment-owned, so this is not a request to set a keymap. It is only about the re-type path not relying on the injected Shift modifier. Happy to open a PR if you would like.

junkerderprovinz:
Fixes #295.

`_handleMobileInput` (the on-screen keyboard / clipboard re-type path) capitalised by injecting `Shift_L` down, then the LOWERCASE letter keysym, then `Shift_L` up, relying on the X server's Shift modifier. On deployments whose X keymap does not bind Shift as a real modifier (for example Xvfb-based bases that ship without a keymap), that Shift never engages and every uppercase letter arrives lowercase, while the physical-keyboard path is fine because it sends the uppercase keysym directly and lets the server resolve the shift level.

This makes the re-type path do the same: send the character's own keysym (an uppercase `A` looks up `XK_A`) and let the server resolve the level, instead of the fragile Shift injection. Lowercase, digits and symbols are unaffected; capitals now keep their case regardless of the deployment keymap.

### DIFF
--- a/addons/selkies-web-core/lib/input.js
+++ b/addons/selkies-web-core/lib/input.js
@@ -2073,22 +2073,15 @@ export class Input {
         // Iterated per codepoint so an astral character (emoji, rare CJK) is sent
         // as one keysym instead of two lone surrogates the server cannot type.
         for (const char of text) {
-            const isUpperCase = char >= 'A' && char <= 'Z';
-            if (isUpperCase) {
-                this.send("kd," + KeyTable.XK_Shift_L);
-                const lowerChar = char.toLowerCase();
-                const letterKeysym = Keysyms.lookup(lowerChar.codePointAt(0));
-                if (letterKeysym) {
-                    this.send("kd," + letterKeysym);
-                    this.send("ku," + letterKeysym);
-                }
-                this.send("ku," + KeyTable.XK_Shift_L);
-            } else {
-                const keysym = Keysyms.lookup(char.codePointAt(0));
-                if (keysym) {
-                    this.send("kd," + keysym);
-                    this.send("ku," + keysym);
-                }
+            // Send the character's own keysym and let the server resolve the shift
+            // level, the same way the physical-keyboard path does. The previous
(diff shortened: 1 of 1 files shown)


junkerderprovinz:
Thanks @ehfd, 4edd73a covers this. `_handleMobileInput` now routes through the shared `_typeText`, which looks up the character's own keysym and sends kd/ku for it with no `Shift_L` injection, so capitals keep their case regardless of whether the X keymap binds Shift. That is exactly what this PR changed, and folding both text paths into one helper (plus the `client_typing_audit.mjs` regression checks) is the better shape, since that is how the two paths drifted apart in the first place. Closing as superseded.

Issue #294, opened 2026-08-07
Mouse acceleration Pointer lock.


thelamer:
Can be AI-driven: Yes, as long as all cases are tested carefully in real-life by people as well as automated tests.
Expected to be workable with: Fable 5, GPT-5.6 Sol, Kimi-K3, Opus 5, or equivalent (risky with Opus 4.8 or under).
Prompt or reference: Comments below.

---

A user brought this up and discord and this is pretty easy to implement. 

When using different browsers they fire relative pointer movement differently and in the case of Firefox very infrequently. 

We might want to implement some kind of multiplication setting here https://github.com/selkies-project/selkies/blob/main/src/selkies/input_handler.py#L3549. Let the users set like between 0.1 and 10.0 to multiply the value by allowing them to slow down or speed up the movement when it pointer lock mode. 

Gives them at least some kind of knob to overcome the browsers shortcomings on pointer lock events.

ehfd:
I think that the best way is to resolve the relative pointer movement behavior in browsers properly, rather than rely on the handle.
The handle is not likely to be used properly by a lot of people.

ehfd:
One-half addressed in https://github.com/selkies-project/selkies/commit/cee0b827de8292f5243a6a60e1f6371298b26c31.

PR #297, opened 2026-08-09
fix(input): request unadjusted movement for pointer lock (#294)

junkerderprovinz:
Relates to #294.

## Problem
Pointer-lock relative mouse movement can feel inconsistent across browsers/OSes. The issue's own suggestion was a manual sensitivity multiplier, but @ehfd's comment on the issue points at the better fix: resolve the browser's relative-movement behavior properly rather than add a user-facing "handle" for people to fight the platform with.

## Fix
`requestPointerLock()` accepts an `{ unadjustedMovement: true }` option (Pointer Lock API v2) that bypasses the OS-level mouse-acceleration curve some platforms apply to `movementX`/`movementY`, giving raw hardware deltas instead. This PR requests it first at both call sites (`Input.js` manual Ctrl+Shift+click lock and `_armPointerLock`'s fullscreen auto-lock), and falls back to a plain `requestPointerLock()` specifically on `NotSupportedError` (the spec-defined rejection when the option isn't available), so engines without support behave exactly as before.

Chromium has supported this since M88; Firefox shipped it in Firefox 152. Older engines that return `undefined` instead of a Promise from `requestPointerLock()` (Safari, Firefox < 122) are unaffected by the extra option and keep working through the existing undefined-vs-Promise duck-typing check already in this code.

## Scope
This is deliberately the low-risk half of the fix. The other, harder half — normalizing Firefox's coalesced/throttled `movementX`/`movementY` event delivery under pointer lock via `PointerEvent.getCoalescedEvents()` — touches a hot input path and needs real cross-browser and real-hardware testing this environment can't provide, so it's left for a separate follow-up rather than bundled in here.

## Verification
`node --check addons/selkies-web-core/lib/input.js` passes. No test harness exists for this addon; this is a source-level review, not a runtime test — happy to adjust based on real browser testing feedback.

### DIFF
--- a/addons/selkies-web-core/lib/input.js
+++ b/addons/selkies-web-core/lib/input.js
@@ -2132,9 +2132,21 @@ export class Input {
             // requestPointerLock() returns undefined (not a Promise) on older
             // engines (Safari, Firefox < 122); failures there surface via the
             // pointerlockerror event instead.
-            const lockPromise = targetElement.requestPointerLock();
+            // Request unadjustedMovement first so relative-move deltas bypass
+            // the OS pointer-acceleration curve; engines that don't support
+            // the option reject with NotSupportedError, so retry without it.
+            const lockPromise = targetElement.requestPointerLock({ unadjustedMovement: true });
             if (lockPromise && typeof lockPromise.catch === 'function') {
-                lockPromise.catch(err => console.error("Pointer lock failed:", err));
+                lockPromise.catch(err => {
+                    if (err && err.name === 'NotSupportedError') {
+                        const fallback = targetElement.requestPointerLock();
+                        if (fallback && typeof fallback.catch === 'function') {
+                            fallback.catch(err2 => console.error("Pointer lock failed:", err2));
+                        }
+                        return;
+                    }
+                    console.error("Pointer lock failed:", err);
+                });
             }
             this.cursorDiv.style.visibility = 'hidden';
             event.preventDefault();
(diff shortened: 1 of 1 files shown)


ehfd:
> The other, harder half — normalizing Firefox's coalesced/throttled movementX/movementY event delivery under pointer lock via PointerEvent.getCoalescedEvents() — touches a hot input path and needs real cross-browser and real-hardware testing this environment can't provide, so it's left for a separate follow-up rather than bundled in here.

My ultimate intervention for this problem does prefer this together.

junkerderprovinz:
Not in this PR. What is here is only `requestPointerLock({ unadjustedMovement: true })` plus a `NotSupportedError` fallback at the two lock sites, +32/-7 in one file, with no change to the movement path.

On bundling it: the coalesced half is not a drop-in on top of this. `_mouseButtonMovement` already routes relative deltas into `_queueCoalescedMouseMove`, which sums `m2` deltas and flushes at most once per animation frame. So if Firefox is only pacing movement differently while still reporting a correctly summed `movementX`, expanding `getCoalescedEvents()` into that path yields the same totals and changes nothing on the wire. If Firefox is actually dropping delta, it does help. Which of the two it is has to be measured on real Firefox and Chromium with real mice before the right change is even known, and the fix then has to touch sampling and pacing in the hottest input path in this file.

That measurement is the part I cannot produce from here, and #294 explicitly conditions this on real-life testing. So I would keep them apart: this one is self-contained, reverts cleanly and can go in on its own, and the sampling rework follows as its own PR with before and after numbers. If you would rather review it as one change, say so and I will fold it into this branch and leave it as a draft until it has been tested on hardware.

ehfd:
Deduplication and test harness addition in https://github.com/selkies-project/selkies/commit/cee0b827de8292f5243a6a60e1f6371298b26c31.

junkerderprovinz:
Thanks for landing it together with the dedup and test harness, that's a cleaner outcome than two separate PRs would have been. Closing this one since the fix is already on main via cee0b827.

ehfd:
@junkerderprovinz I recommend screening the PRs using AI, which would read either AGENTS.md or CLAUDE.md for stylistic and functional decisions.

junkerderprovinz:
That is exactly the workflow already in place here, the target repo's AGENTS.md/CLAUDE.md gets read before every contribution (this repo's own CLAUDE.md is what shaped the self-contained, revert-cleanly approach on this PR). Appreciate you calling it out explicitly.

PR #299, opened 2026-08-12
stop nvml spam, swap to crf as default, revert icon loading to png

thelamer:
Given the wide deployment CRF should still be default for websockets mode. 

NVML does not need to be checked for on repeat so just fail fast. 

On our images we extensively catalogue and only use PNGs for the icon, adding the svg as default breaks this for us. 

I was tempted to just restore the wtype fallback stuff for clipboard, but in general KDE is busted for input and clipboard interactions, the logic removed from input_handler needs to be re-added. To address this under the new design format ext_data_control_manager_v1 needs to be implemented in dcclient.rs in pixelflux and plumbed into input_handler for use on non unicode characters. Just in general I would diff out input_handler on lsio branch and main there is alot of crap in there but a lot of it is for very specific reasons through iterative development with native speakers.

### DIFF
--- a/addons/selkies-dashboard-wish/src/components/dashboard/settings.tsx
+++ b/addons/selkies-dashboard-wish/src/components/dashboard/settings.tsx
@@ -132,17 +132,8 @@ const STREAMING_MODES = [STREAM_MODE_WEBSOCKETS, STREAM_MODE_WEBRTC];
 const DEFAULT_STREAM_MODE = STREAM_MODE_WEBSOCKETS;
 
 const rateControlOptions = ["cbr", "crf"];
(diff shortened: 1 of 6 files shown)


ehfd:
I will review all of this. Thanks.

ehfd:
Incorporated in https://github.com/selkies-project/selkies/commit/d86bbfe948313b851cb240bbda5d99fa66fd0184.

ehfd:
Latter stages implemented in https://github.com/selkies-project/selkies/commit/f985dd486be3f320f2a75d268202f084e06014c3 and https://github.com/selkies-project/pixelflux/commit/938bce5e587e740eaf00940ccb5a9a5f708be830.

PR #301, opened 2026-08-17
fix(input): send the character's own keysym for pasted/typed uppercase (lsio)

junkerderprovinz:
**Reference the issue numbers and reviewers**

@ehfd @thelamer — re-applies my own #296 (issue #295) to the `lsio` branch, which has neither #296 nor its later `main`-branch form (`4edd73a`). Downstream report this is blocking: [junkerderprovinz/krusader#27](https://github.com/junkerderprovinz/krusader/issues/27).

**Explain relevant issues and how this pull request solves them**

`_handleMobileInput` in `addons/selkies-web-core/lib/input.js` wraps uppercase letters in explicit `Shift_L` down/up around a LOWERCASE keysym instead of sending the uppercase character's own keysym. On `lsio`'s server side this doesn't work: `input_handler.py` routes alpha keysyms through `pynput`, while `Shift_L` goes through a separate `xdotool` path with no shared modifier state, so the held Shift is silently dropped and every pasted/typed uppercase character arrives lowercase. This is exactly what #295/#296 originally fixed on `main`, and what `4edd73a` later folded into a bigger commit — neither has reached `lsio`.

**Describe the changes in code and its dependencies and justify that they work as intended after testing**

`_handleMobileInput` now looks up and sends each character's own keysym directly — exactly what its own (unmodified) non-uppercase branch, and this file's other two text-injection paths (`_handleTextInput`, `_updateCompositionText`), already do on `lsio`. Also removes a stuck-Shift hazard: the old raw `Shift_L` send bypassed `_keyDownList` bookkeeping, so a lost release left Shift held with nothing tracking it.

**Not live-tested end-to-end in a running browser session** — I don't have a build/runtime environment for this repo set up. What I did verify:
- Direct comparison to this file's two other (unmodified, already-correct) text-injection paths.
- A full trace through `lsio`'s actual `input_handler.py` confirming the mechanism above.
- A brute-force equivalence check (all non-`A-Z` UTF-16 code units) that the new unconditional loop body is behaviorally identical to the untouched half of the old branch.
- `node --check` on the resulting file.

Happy to do a real browser test if there's a quick way to spin up a dev build — just didn't want to hold the PR on that.

**Describe alternatives you've considered**

[...]

PR #302, opened 2026-08-17
clipboard: sync local clipboard on Ctrl/Cmd+V for Firefox and Safari

junkerderprovinz:
On `lsio`, `window.addEventListener('focus', ...)` in `selkies-ws-core.js` is the only trigger for reading the local browser clipboard and pushing it into the remote streamed session. `navigator.clipboard.readText()`/`.read()` require real transient user activation to succeed without a permission prompt; a `focus` event carries none. Chromium grants the read anyway; Firefox and Safari do not, so on those engines the read silently fails or pops an ephemeral "Paste" prompt on every window focus. This is documented in this repo's own FAQ (added via PR #224) and referenced in issues #258 and #234.

This PR adds a second trigger for the same existing read-and-send logic: a real Ctrl/Cmd+V keydown on non-Chromium browsers, which does carry transient activation. The focus listener stays unchanged for Chromium.

### Prior context

I opened PR #276 for this exact idea against `main` in July. The maintainer noted it was already covered by PR #254's larger merge. That merge, and `main`'s newer modular clipboard handling (`addons/selkies-web-core/lib/clipboard-sync.js`), never landed on `lsio` — `lsio`'s clipboard code is still the older, monolithic `selkies-ws-core.js` layout this PR targets directly.

### Scope

`main` separately carries commit `00ce7394046eb0b89f2bd1892492a348a23db780` ("Performance optimizations"), which contains this same Ctrl+V mechanism plus a much larger, unrelated rework: a video-pipeline change and a new clipboard *write* direction for Ctrl+C needing a new client/server request-response protocol not present on `lsio`. That commit does not apply to `lsio` cleanly (conflicts across roughly a dozen files) and is out of scope here. This PR only carries the read/paste direction, and I traced `sendClipboardData` and `isChromium` to confirm it has no dependency on the unported protocol pieces.

Two small deliberate differences from the upstream mechanism this was adapted from: the key check uses `event.code === 'KeyV'` (physical key position) instead of `event.key`, so it still works on non-QWERTY layouts; and an `event.repeat` guard avoids re-reading the clipboard on every repeat while the keys are held.

### Testing

Verified by static analysis and code tracing: the diff applies cleanly to current `lsio` HEAD, `node --check` passes, and I confirmed no dependency on the unported protocol pieces. **Not tested live end-to-end in a browser.** One caveat inherited from the mechanism this is adapted from: the keystroke itself reaches the remote session synchronously while the clipboard read settles asynchronously, so in principle the very first Ctrl+V after a local copy could deliver the previous clipboard content before the new one lands — a pre-existing characteristic, not something this PR introduces.

### DIFF
--- a/addons/selkies-web-core/selkies-ws-core.js
+++ b/addons/selkies-web-core/selkies-ws-core.js
@@ -2246,7 +2246,10 @@ document.addEventListener('DOMContentLoaded', () => {
     window.location.pathname.lastIndexOf('/') + 1
   );
 
-  window.addEventListener('focus', async () => {
+  // Read the local clipboard and forward it to the remote session. Runs on window
+  // focus (all browsers) and, for non-Chromium engines, on a real Ctrl/Cmd+V
+  // keydown as well -- see the keydown listener below for why.
+  async function readLocalClipboardAndSend() {
     if (isSharedMode || !window.clipboard_enabled || !clipboard_in_enabled) return;
 
     if (!enable_binary_clipboard) {
@@ -2255,11 +2258,11 @@ document.addEventListener('DOMContentLoaded', () => {
         .then((text) => {
           if (!text) return;
           sendClipboardData(text);
-          console.log("Sent clipboard text on focus via sendClipboardData");
+          console.log("Sent clipboard text via sendClipboardData");
         })
         .catch((err) => {
           if (err.name !== 'NotFoundError' && !err.message.includes('not focused')) {
-             console.warn(`Could not read text clipboard on focus: ${err.name} - ${err.message}`);
+             console.warn(`Could not read text clipboard: ${err.name} - ${err.message}`);
           }
         });
     } else {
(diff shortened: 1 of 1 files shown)


thelamer:
lsio is not a permanent thing, the `dev` image tag on both baseimage and webtop run selkies head. 

```
lscr.io/linuxserver/webtop:dev
```

The plan is to transition everything over after heavy testing and selkies side releasing its own desktop images to replace the aging glx and egl desktops.

junkerderprovinz:
Thanks for the context, that is really good to know. Funnily enough I came at this from the other side today: you suggested the same `dev` BASE_TAG switch on one of my own images over at junkerderprovinz/krusader#27, so I moved it across and it has been working nicely. Clean drop-in on the same Ubuntu Resolute series, and the clipboard fix is genuinely in the shipped bundle. A small real world vote of confidence for the direction, at least.

On #301 and #302, both target `lsio` specifically, so my instinct is to leave them open while lsio-pinned tags are still in active use during the transition you describe. No pressure either way though, entirely your and LSIO's call if you would rather close them and let `dev` carry it.

thelamer:
Right now lsio selkies is pinned unless there is a major bug that needs to be fixed it will be where it is at until the transition to main selkies and selkies build pixelflux/pcmflux. But your testing on this image with your userbase is valuable to fix any issues. It is a direct drop in replacement that is the purpose of the dev tag right now.

ehfd:
OK, let's focus on delivering v2.0.0 as soon as possible, then.

PR #305, opened 2026-08-18
fix: Various QA fixes

ehfd:
## Summary

This change closes a set of reliability and parity gaps found by empirical testing across the
transport × backend matrix, and ships the container pieces the dashboards' apps panel depends on.
Every fix below is pinned by a new or extended test, and the measurements quoted were taken on the
included suites (local shaped-loopback runs, plus a bare-metal 2×P100 node with real `tc` shaping
for ground truth).

## File transfers no longer fight the stream

New `TransferPacer` shared by the files API in both directions:

- **Adaptive mode (default)**: downloads are held inside a shared allowance driven by the download
  socket's unsent-queue depth (`SIOCOUTQNSD`), falling back to TCP RTT inflation where the ioctl is
  unavailable. AIMD with a per-epoch recovery ceiling; a connection offering no gauge passes
  unpaced rather than being throttled blindly. The RTT floor is per-connection so a nearby client
  cannot make a distant one read as permanently congested.
- **Static cap**: `SELKIES_FILE_TRANSFER_LIMIT_MBPS` bounds both directions from one shared bucket;
  negative/non-finite values resolve to 0 (pacing off). Upload reads have no honest server-side
  gauge (the queue stands in the client's kernel), so they ride the static-cap leg only.
- Measured on a contested 6 Mbit/s link with the stream running: the download completes while the
  stream holds a ≥30 fps median with bounded stalls; static caps land within the configured rate
  (2.8–2.9 Mbit/s against a 3 Mbit/s cap on the GPU node under `tc`). Without any cap, transfers
  saturate the link as before — pacing only engages when configured or when the gauge sees queueing.

## WebRTC datachannel stays responsive under large payloads

Gzipped datachannel messages now inflate concurrently and re-enter an ordered queue only for
order-sensitive kinds (clipboard transfers, `server_settings`); everything else routes immediately.
Input echo gap under a 24 MB clipboard decode: **3.4–15 ms → ≤0.2 ms**.

## App commands: failures surface, optimistic UI rolls back

- The server now watches spawned client commands to completion. A launch failure or nonzero exit
  is reported as a `command_error` system action carrying the exit status, runtime, and the echoed
  command (126/127 get a "command not found" hint). Output stays discarded so a long-running app
  can never block on a full pipe.
- Over WebRTC the notice targets the requesting peer's channel (broadcast fallback if that peer
  reconnected); over WebSockets the requesting socket is notified in its own cmd branch.
- Both dashboards share a new `appCommands.js` contract (previously divergent copies): posted
  install/remove/update/launch commands are tracked, and a matching `command_error` settles the
  pending entry — rolling back the optimistic installed-apps update, or suppressing a stale launch
  exit as application lifecycle rather than a launch failure. When the server cannot run commands
  at all (`SELKIES_COMMAND_ENABLED=false`), the apps UI disables actions and says why.
- New strings translated across all eight locales.

## One encoder knob drives both transports

`encoder_rtc` is gone; `SELKIES_ENCODER` / `--encoder=` now drives WebSockets and WebRTC alike.
In WebRTC mode the published menu is filtered to what WebRTC can produce, and a websockets-only
choice (`jpeg`, `h264enc-striped`) falls back with a logged warning. The operator's menu and
value — and a client's own pick, tracked separately — are restored on the switch back, so nothing
is lost across a transport round trip. Docs, per-display key routing, and both dashboards'
settings panels updated to the single knob.

## Auth changes under a live page recover cleanly

New shared `auth-guard.js` (both cores, both dashboards): any same-origin 401 reloads the page
once into the browser's credential prompt instead of reconnecting forever against rejected
websockets; a loop-breaker caps how many reloads a broken credential can cause. Covered by a new
e2e suite that flips Basic auth on under a connected page (reload storm observed: exactly 1).

## IME composition is correct on every platform branch

The commit `textInput` listener is no longer platform-gated: real IBus Chrome delivers the commit
as `textInput` after an empty-data `compositionend`, and a handler that only listens on Linux
erased every committed syllable sent from Windows/macOS ("keeps replacing one symbol"). The
CDP/insertText anatomy (commit present in both events) deduplicates instead of typing twice.
A new suite replays both anatomies — including mid-syllable preedit probes — across Chromium
(Linux/Windows/macOS-spoofed) and WebKit in both event orders: 19/19.

## Clipboard robustness

- Inbound clipboard images are re-encoded by the browser on write, which made the focus read-back
  see "new" content and echo once; the re-encoded form is now remembered under the synced value's
  signature and suppressed.
- Reads flush behind any just-landed server push so the pre-push content can't bounce back as a
  stale change.
- Wish dashboard: re-picking the same file fires again, and core-emitted clipboard skips surface
  on the same warning channel as the classic dashboard.

## Primary resolution pinning respected client-side

With `enable_resize=false` the client no longer requests resizes for the primary display nor
restyles the layout onto a size the server will ignore; secondary displays keep their independent
resize path (matching the server's gate).

## X11 reported DPI matches the rendered DPI

`set_dpi` now also stamps the root pane's physical size (RandR) so `xdpyinfo`-style consumers see
the DPI the desktops were told to render at, idempotently (a matching size posts nothing, keeping
repeated settings payloads ConfigureNotify-free). Resizes keep sizing monitors in mm at the
applied density instead of flipping back to 96, and `selkies-primary` carries the RandR primary
flag so panels tile against it.

## Example container

- **proot-apps** ships in the image with a `/selkies-proot` wrapper (seeds/refreshes the per-user
  install from `/opt/proot-apps`), `st` as the terminal the apps panel launches, and
  `SELKIES_COMMAND_ENABLED=true` so the panel works out of the box.
- setuid helpers (chrome-sandbox, polkit, `sudo-root`) are restored to root ownership after the
  rootless fakeroot build layers: fakeroot swallows a setuid chmod even at uid 0, which left
  Chrome aborting at launch wherever unprivileged user namespaces are unavailable.
- On Wayland without a render node, Xwayland is wrapped to pin the Mesa EGL vendor ICD
  (`WLR_XWAYLAND`), fixing a startup crash through the NVIDIA GBM stack on GPU nodes that expose
  no `/dev/dri` render device.
- LXQt session: `XDG_MENU_PREFIX` exported on both backends (empty launcher menu otherwise), and
  the icon theme falls back to breeze when the configured theme lacks fullcolor glyphs (blank
  volume plugin). labwc titlebars gain the breeze icon set and window-icon layout.
- `PIPEWIRE_LATENCY` becomes an overridable default (`256/48000`).

## Tests

Five new suites plus harness upgrades, all registered in the tier table:

| Suite | Tier | Checks |
|---|---|---|
| `unit/test_transfer_pacer.py` | unit | pacing-rate, AIMD-epoch, gaugeless and upload contracts |
| `e2e/test_auth_refresh.py` | e2e | 401 reload + storm cap |
| `e2e/test_ime_composition.py` | e2e | 19 across both anatomies × platforms × WebKit orders |
| `perf/test_transfer_saturation.py` | perf | 11 on a shaped loopback link (`tests/tools/bwrelay.py`) |
| `tests/tools/bwrelay.py` | tool | token-bucket relay with bounded receive buffers |

The browser harness is engine-parameterized (system Firefox and WebKit alongside Chromium).
Full validation on this tree: transport × backend matrix **78/78**, dashboards **19/19** (both
bundles rebuilt), IME **19/19**, auth-refresh **3/3**, saturation **11/11**, unit tier **18
suites**; GPU-node runs under real `tc` shaping confirm the pacer's equilibrium numbers.

New and touched code follows the Google-docstring + 3.9-safe type-hint standard.


ehfd:
@junkerderprovinz @thelamer Would like some tests on this PR to fix bugs in Selkies. This is bugfix-only, not a feature.

ehfd:
@PMohanJ @vishalkadam47

junkerderprovinz:
TL;DR: built a dedicated GPU-accelerated test box just for this kind of testing and ran the full claimed e2e/perf suite for real. Unit, auth-refresh and IME match exactly. Matrix is 69/78 (Wayland-only, traced to missing pixelflux methods, not this PR's fault). Saturation is 10/11 (stream degrades under a throttled upload). Gamepad release looks solid on this PR, worse on the parent.

Since the e2e/perf tiers need a real browser and GPU, and I wanted to give this an actual thorough pass rather than a code read, I set up a dedicated GPU-accelerated sandbox on my own hardware (real NVIDIA GPU, Firefox and Chrome, full desktop) just to run this kind of test properly, and I'll keep it around for future PRs like this one. Ran everything the PR claims numbers for.

**Unit, auth-refresh, IME:** all match exactly, 18/18, 3/3, 19/19.

**Transport×backend matrix:** 69/78, not 78/78. Both X11 backends are perfect (`ws-x11` 18/18, `wr-x11` 20/20). All 9 failures are on Wayland (`ws-wl` 14/19, `wr-wl` 17/21), clustered on clipboard sync and secondary-display support. Traced it to `input_handler.py` calling four pixelflux methods that don't exist anywhere in pixelflux, including current main: `set_keymap_overlay`, `clipboard_write_app`, `list_outputs`, `create_output`. Not something this PR introduced, looks like the Wayland work targets a pixelflux version that hasn't shipped yet.

**Dashboards:** 19/19 is real with `test_dashboards.py all`, worth knowing that `tests/suites.py` (what CI's own e2e run executes) only registers 15 of them, the other 4 sit in an unregistered `gates` block.

**Saturation:** 10/11. `upload-static` wants median fps >=45 and stall_frac <=0.10, measured 34 and 0.16, the video stream visibly degrades while an upload is paced to its cap. Possibly connected to the TransferPacer epoch bug from my last comment, haven't proven the link yet.

**Gamepad release:** ran this one on both the PR and its parent commit back to back on real hardware. This PR: 3/4, only the WebRTC peer-disconnect case times out. The parent: 0/4, worse across the board. So no regression here, if anything the opposite.

**Three more, unrelated to this diff:** `test_protocol.py`, `test_parity_checks.py`, `test_display_seed.py` all hit `RuntimeError: no current event loop` on Python 3.14 (bare `asyncio.get_event_loop()` calls). CI never catches this since the integration/e2e tier pins Python 3.12, only the unit tier matrix-tests 3.9/3.14.

Didn't run the suites the PR body doesn't put numbers on (scroll, browsers, clipboard_large, cross_transition, two_displays, gamepad_uinput, dpi_accuracy, regressions, software_decode, soak), stopped at what was actually claimed and verified all of it.

ehfd:
@junkerderprovinz Thank you for the tests!!!

What each finding turned out to be, and what changed:

- Wayland matrix 69/78 — the reviewer's box ran PyPI pixelflux 2.0.0; I verified its wheel lacks ten methods selkies drives (their four plus six more). The methods do exist on pixelflux main — as Rust #[pymethods], which their Python-side grep missed. Selkies now logs one startup warning naming exactly what the installed pixelflux lacks, and the matrix prints a PREFLIGHT line before Wayland blocks so this self-diagnoses. Action only you can take: ship pixelflux 2.1.0 to PyPI, then pin pixelflux>=2.1.0 in pyproject (I left it unpinned so installs don't break before the release).
- TransferPacer epoch bug — their earlier comment is nowhere on the PR (likely a pending review only they can see; the reply asks them to submit it). I found a real one regardless: a clear sample inside the post-cut drain window closed the epoch, so an oscillating gauge re-armed the recovery ceiling from each freshly cut rate — ratcheting toward the floor. Fixed; the new unit check fails 12/14 on the old code, 14/14 on the fix.
- Saturation upload-static (their fps 34) — the harness, not the server: bwrelay's bucket had the same double-credit bug the server pacer already guards against, giving bulk flows ~2× the rate and parking the balance ≤0 where small takes starve. An isolation probe showed a 32-byte echo getting one response in 8 s through the shipped relay beside an upload; after the fix (honest deficit accounting, FIFO-fair takes, 8 KiB grain) it reads p50 3.6 ms / p95 4.6 ms. Suite now 11/11 with upload-static at 60 fps / 0.00 stalls — matching the re-run tc ground truth on the P100 (2.84 Mbit/s paced, 60 fps / 0% throughout), which proves server-side pacing was already right on a real link. upload-open gained an upper bound so the accounting bug can't silently return.
- Gamepad WebRTC release timeout — reproduced here too (not box-specific): ICE consent (6×5 s) was the only WR death detector. Fix is input-level liveness: a js,h held-state heartbeat from the client (same idiom as the kh key heartbeat) plus a 2 s server sweep. Release suite extended with SIGKILL variants: 6/8 → 8/8, WebRTC release >30 s → 2.0 s on both death modes.
- Python 3.14 event-loop errors — the three tests use asyncio.run() now, validated on 3.12 and on a real 3.14.6 env with cp314 pixelflux/pcmflux wheels I built (11/11, 2/2, 1/1 with the server itself on 3.14).
- Unregistered dashboards gates — suites.py now carries the gates selector; CI runs all 19.

Final validation: unit 18/18 suites, matrix 78/78, dashboards 19/19 (including the CI SELKIES_INJECT=1 bundle variant), IME 19/19, auth-refresh 3/3, saturation 11/11, gamepad 8/8, integration trio on 3.12 + 3.14, and the four P100 tc scenarios all at 60 fps / 0% stalls — with down-cc-6 at 3.91 Mbit/s, identical to before the epoch fix, so the adaptive equilibrium paid nothing for it.

PR #307, opened 2026-08-19
fix: Pointer lock improvements

### LINKED ISSUE #294: Mouse acceleration Pointer lock.

ehfd:
Closes https://github.com/selkies-project/selkies/issues/294.

Proposed solution to https://github.com/selkies-project/selkies/issues/294.

@junkerderprovinz Appreciate if this can be reviewed and tested.

@thelamer


thelamer (approved):
Tested gzdoom at scale and 800x600 with pointer lock no more float. Works normally outside of scaled or manual resolution modes.

ehfd:
Let's wait for @junkerderprovinz then merge.

PR #308, opened 2026-08-19
Webcam

thelamer:
Test image
```
docker run --rm -it --shm-size=1gb -p 3001:3001 taisun/random-images:webcam-kde bash
```

This works similarly to the gamepad interposer, tested with multiple applications the big one being chromium web browsers. Firefox. ffplay, vlc, mpv. 

OBS stands out as one that does not work, it integrates with v4l2 at a lower level than our interposer can handle. 

Clients tested: 

Desktop: Chromium based, Firefox
Mobile: Android chrome, iPad safari


ehfd:
https://www.ideasonboard.com/news/pipewire-is-the-new-v4l2loopback/

What does the interposer do compared to pipewire-v4l2, considering that you weren't able to get OBS working yet?

thelamer:
> https://www.ideasonboard.com/news/pipewire-is-the-new-v4l2loopback/
> 
> What does the interposer do compared to pipewire-v4l2, considering that you weren't able to get OBS working yet?

why-not-both-why-not-4228684839

This has nothing to do with pipewire it just provides an optional interposer and default flagged false webcam setting to the codebase. You can always add multiple backends in the future fed by the webcam frames and not all Linux applications will support pipewire.

ehfd:
@thelamer I'll investigate more, but the point is that pipewire-v4l2 is literally an interposer that does `LD_PRELOAD` to remove the need for `v4l2loopback`.

The investigation is on me. I'll get back.

thelamer:
Added support for obs, this covers every program I could think of testing and we now shim glibc syscalls for /dev/video0 and /sys/class/video4linux.

ehfd:
@thelamer I am cautious about merging this right now due to: (1) Having only MJPEG as the codec and interfacing in Python (can't encode this codec with hardware from the browser), (2) Not integrated with pixelflux, which is the right place to interface this, just like pcmflux, where the microphone interface is handled.

I'll wait a bit and sort other things out before deciding. The footprint is relatively benign and trivial to rebase.

thelamer:
Head image has been updated 
```
docker pull taisun/random-images:webcam-kde
docker run --rm -it --shm-size=1gb -p 3001:3001 taisun/random-images:webcam-kde bash
```

thelamer:
Added OBS to test image for testing.

thelamer:
Off tabbing has a heavy effect on the webcam encoder while the webcam is enabled, best solution I found was to stop feeding it when off tab. Without this you  get permanent desync in the webcam stream. We might want to implement something like this for mic as well. It is just the nature of browsers sleeping workers when off tab.

ehfd (approved):

PR #326, opened 2026-09-05
feat: Stream H.265, VP8, VP9 and AV1 over WebSockets and WebRTC

### LINKED ISSUE #30: Add support for X11 zero-copy GPU screen capture (NVIDIA NvFBC)
Can be AI-driven: Yes, as long as all cases are tested carefully in real-life by people as well as automated tests. However, **expert guidance is required due to working with X11 intrinsics**.
Expected to be workable with: Fable 5, GPT-5.6 Sol, or equivalent (max thinking, requires expert guidance, especially with Opus 5, Kimi-K3, or under).

NOTE that we are using pixelflux over GSTreamer now.

---

Addenum by @ehfd :

- NvFBC is still something we want.
- We shall not use any variant of kmsgrab, since that requires CAP_SYS_ADMIN, which is just short of root privileges in the host. Otherwise, we will have to live with ximagesrc (which uses XShmGetImage under it) for general-purpose usage, and port to pipewiresrc (#46) in the future, when there are blueprints for Wayland support.

### LINKED ISSUE #34: [META] Codec improvement and additional support including AV1/H.265, Intel/AMD GPU encoding support
Can be AI-driven: Yes, as long as all cases are tested carefully in real-life by people as well as automated tests. However, **expert guidance is required due to working with a deep understanding of codecs and protocols in general**.
Expected to be workable with: Fable 5, GPT-5.6 Sol, or equivalent (max thinking, requires expert guidance, especially with Opus 5, Kimi-K3, or under).

**H.265, VP8, VP9, AV1 support is needed. This is now on Pixelflux, not GStreamer.**

---

Looking for contributors to help out.

Guide for improving video encoder parameters (also applicable to non-NVIDIA encoders): https://docs.nvidia.com/video-technologies/video-codec-sdk/12.2/nvenc-video-encoder-api-prog-guide/index.html#recommended-nvenc-settings__table-nvenc-settings

**Remember that live (as in < 100 ms) streaming is all about keeping the bitrate consistent at all frames to minimize congestion from unexpected bandwidth spikes (which lead to stutters and latency spikes) and keeping the buffer small. Use the configured Constant Bitrate bandwidth as possible at all times but do not go over it (ideal maximum overshoot tolerance = 10%).**

**Priority: Latency > Quality (minimum sacrifices of <= 5% in latency for quality or bandwidth consistency is acceptable, using more GPU resources or CPU cores also acceptable if without latency impact) >> Overall Bandwidth Efficiency (since encoded frames are only used once unlike .mp4/.mkv)**

**This is also applicable to other encoders if available.**

- Ultra-low latency or low latency Tuning Info (**with zero-latency/zero-reorder-delay/low-latency modes**)
- Rate control mode = CBR
- Multi Pass – Quarter/Full (evaluate and decide) (**recommended for hardware encoders if available, might backfire with software encoders**)
- Very low VBV (=CPB) buffer size (e.g. single frame = bitrate/framerate) (**directly correlates with web browser decoder jitterbuffer latency but causes periodic blur with keyframe/GOP when too low, 1x in HW and 1.5x in SW recommended with infinite keyframe/GOP, otherwise increase n (2-3) times**)
- Unidirectional B Frames (**incompatible with WebRTC in current (2023) web browsers**)
- Infinite GOP length (**prevents periodic blurring and bandwidth overshooting**)
- Adaptive quantization (AQ) enabled** (**may lead to mixed results**)
- Long term reference pictures*** - https://docs.nvidia.com/video-technologies/video-codec-sdk/12.2/nvenc-video-encoder-api-prog-guide/index.html#long-term-reference (**may lead to mixed results**)
- Intra refresh*** - https://docs.nvidia.com/video-technologies/video-codec-sdk/12.2/nvenc-video-encoder-api-prog-guide/index.html#error-resiliency-features__section_klf_nr1_nlb (**was not compatible in WebRTC with x264, need to revisit**)
- Non-reference P frames*** (**collides with congestion control for some reason**)
- Force IDR*** (**implemented using WebRTC PLI**)

***: These features are useful for error recovery during transmission across noisy mediums. For usecases where the client requires reduced video memory footprint (**this typically means reducing any buffer that exists, preventing frame lag, except the ones unrelated to latency, stated below**), following guidelines should be followed.

- Avoid using B-frames. B-frames requires additional buffers for reordering, hence avoiding B-frames would result to savings in video memory usage. (**Incompatible with WebRTC in current (2023) web browsers**)
- Reducing the maximum number of reference frames doesn't matter. Reducing number of maximum reference frames results in NVIDIA display driver allocating lesser number of buffers internally thereby reducing video memory footprint. (**May impact latency, but long reference should be safe**)
- Single pass rate control modes doesn't matter. Two pass rate control consume additional video memory in comparison to single pass due to additional allocations for first pass encoding. Two pass rate control mode with first pass with full resolution consumes more than first pass with quarter resolution. (**Not influential to latency but adds GPU load, two-pass is recommended nonetheless**)
- Adaptive Quantization / Weighted Prediction has unknown effects. Features such as Adaptive Quantization / Weighted Prediction allocate additional buffers in video memory. These allocations can be avoided if these features are not used. (**Not influential to latency but adds GPU load, so worth checking both features**)
- Avoid Lookahead. Lookahead allocates additional buffers in video memory for frames that are buffered in the lookahead queue. (**Important because it holds generated frames**)
- Avoid Temporal Filter. Temporal filter requires neighbouring frames and allocates additional buffers in the video memory. (**Important because it holds generated frames**)
- Avoid UHQ Tuning Info. UHQ Tuning Info enables lookahead and temporal Filter, that have higher memory requirements. (**Important because it holds generated frames**)

- It should be possible to obtain the sweet spot of encoder maximum and minimum QP parameters

https://multi.app/blog/making-illegible-slow-webrtc-screenshare-legible-and-fast
https://multi.app/blog/measuring-shared-control-latency

Note: HEVC is low hanging fruit and makes a massive difference, 444 HEVC encoding is supported on almost every intel chip and allows zero copy 444. Plus 10 bit greatly increases gradient fidelity.

ehfd:
Closes #34
Closes #30

Implements #34 with selkies-project/pixelflux#6 (pixelflux 2.1.0, pinned together). pcmflux is unchanged.

## Encoders
- `--encoder=` gains `h265enc`, `vp8enc`, `vp9enc` and `av1enc` next to `h264enc`, `h264enc-striped` and `jpeg`; each runs on NVENC or VA-API where the GPU carries the codec, else on the software encoder pixelflux was built with (`pixelflux.SOFTWARE_ENCODERS`), on X11 and Wayland alike. Rate control stays CRF or CBR for every codec.
- The web client reads each stream's codec from the wire header and its decoder string from the key frame (`lib/wire-codecs.js`: avc1/hev1/vp8/vp09/av01), probes `VideoDecoder.isConfigSupported` per codec to filter the menus, and steps a refusal ladder the server confirms: a codec the engine declines goes to `h264enc` (JPEG when H.264 itself is refused), a held encoder is reported once. The video worker takes the shared codec helpers as raw module source, which survives bundling.
- Full colour (4:4:4) follows the codec (H.264 and H.265) and the engine's answer for it.

## WebRTC
- The vendored RTP stack packetizes H.265 (RFC 7798), VP9 (RFC 9628) and AV1 (AOM v1.0) next to H.264 and VP8, every packer reporting the frame's key-frame flag to the pacer from the bitstream. The webcam uplink takes every one of them: H.265 and VP9 depayload per packet, and an AV1 frame is reassembled from its packets once the jitter buffer holds them all (one OBU's fragments span packets), so the browser may send its camera as AV1.
- The offer puts the display's codec first, H.264 behind it and the rest after: a browser that declines the codec is answered with H.264 and the display moves to `h264enc` (logged), unless the operator holds the encoder; a live encoder change switches each peer's payload type to the codec it negotiated, without renegotiation. The dashboards filter the WebRTC menu by `RTCRtpReceiver.getCapabilities`.

## Merge order
CI installs the capture stack from pixelflux main HEAD, so its unit and integration suites fail on this branch until selkies-project/pixelflux#6 is merged and its pre-release wheels are published. A pixelflux without the codec surface (`SOFTWARE_ENCODERS`, `CaptureSettings.codec`) now stops the server at import with one message naming it, instead of a traceback per client connect. The local runs below used the branch's own wheels.

## Verified
- Unit: `tests/unit/test_codec_encoders.py`, `test_rtp_packetizers.py`, `test_encoder_demotion.py`, `test_rate_control_defaults.py`, `test_track_transform.py`; the unit tier.
- End-to-end on the test host with the GPL wheel (NVENC H.264/H.265, software VP8/VP9/AV1): `test_codecs.py` (every codec, H.264 included) ws-x11 on Chromium, Firefox and WebKit, ws-wl on Chromium, wr-x11 on Chromium (VP8/VP9/AV1 negotiated, H.265 declined and answered with H.264); `test_webcam.py av1` (Chromium's camera sent as AV1, reassembled into the virtual camera at the camera's rate), `webrtcpref` (the camera sender steered to VP9 and AV1 by `webcam_encoder`) and `encoderpref` (VP9 and AV1 pinned over WebSockets on Chromium, their codec ids alone on the wire); `test_encoders.py` ws-x11/wr-x11, `test_software_h264.py`, `test_full_color.py` (all selectors), `test_browsers.py`, `test_dashboards.py`, `test_dashboard_matrix.py` wish-webrtc-x11 (live switch to VP8 over WebRTC), `test_webcam.py` webrtc, `test_protocol.py`, `test_parity_checks.py`, `test_display_rate_control.py`, `test_capture_api.py`, `test_capture_api_extra.py`, `test_wayland_seam.py`.
- The same with the non-GPL wheel (`PIXELFLUX_ENABLE_GPL=0`: OpenH264, kvazaar, libvpx, SVT-AV1): `test_codecs.py` on every engine and transport, `test_software_h264.py` x11 and wl, `test_encoders.py`, `test_full_color.py` (the locked and pinned scenarios read the build: OpenH264 streams a locked full colour as 4:2:0), `test_browsers.py`, `test_capture_api.py`, `test_capture_api_extra.py`, and every software encoder of that build decoded in a browser with software encoding forced.
- Playwright's WebKit answers `isConfigSupported` yes for AV1 and then refuses every AV1 key frame at decode (libaom- and SVT-AV1-encoded alike; the system GStreamer `av1dec` decodes the same bytes). The client treats a decode-time refusal of a codec with a rung below it as a refusal, steps to `h264enc` without the crash-count reload, and hands the next stream to the video worker while it drops the refused stream's stragglers; the codec suite holds such an engine to the refusal outcome.
- The same WebKit refuses H.264 in Annex B form and decodes it with an `avcC` description and length-prefixed NAL units (x264 and NVENC streams alike; Chromium and Firefox take Annex B). The client now asks its decoder once at load, in a worker, with a 43-byte 16x16 key frame, keeps Annex B wherever it decodes, and only where it is refused builds the description from each key frame's SPS and PPS and reframes the chunks, on the page and in the worker, full-frame and striped. H.264 joined the codec suite, so WebKit's H.264 picture is sampled from now on; it had never been.
- The VP8, VP9 and AV1 quantizer tables ramp to the codec's maximum at the top of the quality index instead of sitting on it from 42 up, so every step of `video_crf` still moves the quantizer.
- The webcam uplink offers the same codecs on both transports. Over WebSockets the measured WebCodecs ladder gained VP9, AV1 and H.265 behind H.264 and VP8, and `webcam_encoder` takes those names; over WebRTC the same setting sets the camera sender's codec among the ones the answer negotiated (the per-encoding `codec` parameter, since a browser answering an offer sends with the offer's first codec whatever its own answer lists first). Both dashboards list the new values.
- A capture whose codec no encoder could serve streams H.264 after pixelflux's selection ladder demotes it; the server now reads the codec the capture streams once frames flow (`ScreenCapture.active_codec()`), moves the display's setting to that encoder, switches its WebRTC senders and re-announces the settings, so clients are told the encoder they really receive.

## X11 zero-copy GPU capture (NvFBC)

Closes #30, implemented in selkies-project/pixelflux#6. A session on an NVIDIA GPU encoding with NVENC now captures X11 without copying a frame: the NVIDIA X driver composites the screen into video memory and the encoder reads that buffer in place, which is what the Wayland dmabuf path already did, so the two backends are finally at parity on this. Measured at 1920x1080 on a Tesla V100: 2.53 ms per frame against 5.02 ms for the shared-memory path in H.264, and 3.05 against 5.56 in H.265, with the host CPU under 0.2 ms a frame.

Selkies needs no setting for it — the capture library detects it and says in the session log which path it took. The documentation here had gone stale in two places: the component reference still said H.265 and AV1 were "planned but not yet implemented" in the capture path, and neither it nor the design overview said where a frame is copied on each backend.

### An X server with no RandR output

Reaching that capture on a board whose driver reports no display device at all needed one change here. Such a server has no output, so the session's displays could not be published as RandR logical monitors and the desktop had no screen to lay windows out on; the desktop image worked around it by handing the whole session to a software framebuffer server, which costs exactly the GPU capture this closes. The monitor paths now treat the output as optional: a display becomes a monitor carrying no output, the modes and the primary-output flag are skipped where there is nothing to attach them to, and clearing the layout leaves one monitor covering the framebuffer, because such a server has no screen of its own to fall back to.

Verified in the desktop image on a Tesla P100 (a board the driver reports zero display devices for, confirmed on bare metal with nothing running, and which no `ConnectedMonitor`, `UseDisplayDevice` or `CustomEDID` setting can give a head): the desktop paints in full with its panel, a 1280x720 display client resizes the framebuffer and the monitor follows it, the monitor survives the client leaving, 1801 stripes streamed, and NvFBC captures the painted desktop on that outputless screen. Boards from Volta on carry virtual display devices and take the ordinary output path, unchanged.

### Test suites

Two unit suites failed for reasons that predate this branch, both because they reached outside the tier's contract of needing only the source tree, and both are fixed here: the window-manager suite demanded a live X display for its last check (it now skips that half unless one is configured), and the listen-address suite resolved a hostname through the system resolver, which the binder does on the event loop's executor, so a resolver that never answers hung the suite until the 120-second timeout rather than failing it (the resolution failure is now simulated, and the taken-port check holds its port instead of looking one up and asking for it later). The unit tier is 81 passing, 0 failing.

### The colour signal a client is told

With the capture library converting every session with BT.709 (selkies-project/pixelflux#6), the
two places selkies declares a matrix follow it, per codec rather than as one constant:

- `decoderColorSpace` (WebSockets, WebCodecs) gives VP9 `bt709` and keeps VP8 on `smpte170m`.
- The RTP colour-space header extension is now a per-MIME table in `webrtc/rtcrtpsender.py`
  (`RTP_COLOR_SPACE`) rather than a single tuple carried on each sample: the colour signal
  belongs to the codec, not to the frame, so `EncodedPacket` loses its `color_space` field. It
  covers VP8 and VP9 only, as before. Broadening it to every codec was tried and reverted: a
  receiver takes the extension over the bitstream, and H.264, H.265 and AV1 declare their range
  as well, which a 4:4:4 session signals as full and the table has no way to know.

VP8 keeps BT.601 because its keyframe header can name no other matrix: told BT.709 in either
channel, Chromium and WebKit paint it correctly and Firefox reads neither and inverts BT.601,
measured at 21 levels over WebSockets and 23 over WebRTC.

The e2e colour fixture moves from a saturated green block to a red one. Chromium and Firefox both
convert through libyuv, whose default build clamps the BT.709 Cb→B coefficient, which costs a
green block 11 levels of blue and a red one 2 — a wrong matrix still lands more than twenty levels
off either way, so the check keeps its 10-level tolerance and stops tripping over a receiver-side
rounding it cannot fix. `test_codecs.py` waives the matrix check only for VP8 on WebKit now: its
VP9-over-WebRTC waiver is gone, that cell measuring 2 levels since the stream and the extension
agree.

Verified: `e2e:test_codecs` and `e2e:test_encoders` on ws-x11, wr-x11 and ws-wl, the unit tier,
and a chart-and-field sweep of all five codecs on Chromium, Firefox and WebKit over both
transports.


ehfd:
https://git.unom.io/unom/punktfunk

This is Apache/MIT, which means the code can be used without limitations. Can someone use Fable or Astra-tier AI to see performance- or capability-related improvements while the AI ingests https://github.com/selkies-project/pixelflux/pull/6, https://github.com/selkies-project/pixelflux/issues/2, https://github.com/selkies-project/selkies/pull/326, and https://github.com/selkies-project/selkies/issues/34, and then make a PR against the `codecs` branch?

Reference the prompt in https://github.com/selkies-project/pixelflux/issues/2#issuecomment-5557605269 as well.

PR #333, opened 2026-09-06
fix(input): keep macOS on its own pointer acceleration under lock

### LINKED ISSUE #328: [BUG] Raw unadjustedMovement Pointer Lock causes sluggish/heavy mouse movement on macOS and should not be enabled by default
Describe the bug

When using Selkies Game Mode — fullscreen with pointer and keyboard locked, mouse movement on my macOS client feels significantly sluggish, heavy, and unnatural.

Selkies requests Pointer Lock using unadjustedMovement: true, effectively:

element.requestPointerLock({
    unadjustedMovement: true
});

I found that changing Selkies to use standard Pointer Lock:

element.requestPointerLock();

dramatically improves the mouse behaviour.

This does not necessarily appear to be a defect in the Pointer Lock API itself. unadjustedMovement is an optional raw-input mode intended to bypass operating-system mouse acceleration. This can be useful for applications such as FPS-style camera control, but it is not necessarily desirable for strategy games, older games, desktop-style cursors, or other applications streamed through Selkies.

Because Selkies Game Mode can be used with many different applications, enabling raw movement by default can result in degraded pointer behaviour for some clients, particularly macOS.

Reproduction / isolation

To isolate the issue, I created a minimal local HTML/JavaScript Pointer Lock test on the same macOS.

This removes the complete Selkies/server stack from the test:

Physical mouse
    ↓
macOS
    ↓
Browser
    ↓
Pointer Lock API

There is:

No Selkies server
No network streaming
No Pixelflux
No Wayland compositor
No Wine
No remote game

I compared the following two modes.

Test A — standard Pointer Lock
element.requestPointerLock();

Result:

Mouse movement feels natural and responsive.

Test B — raw/unadjusted Pointer Lock
element.requestPointerLock({
    unadjustedMovement: true
});

Result:

Mouse movement immediately feels sluggish, heavy, and unnatural.

The difference is very noticeable when switching between the two modes.

Selkies Game Mode behaviour

When activating Game Mode — fullscreen with pointer and keyboard locked, the Selkies web client contains logic equivalent to:

let promise = this._unadjustedMovement
    ? element.requestPointerLock({ unadjustedMovement: true })
    : element.requestPointerLock();

The _unadjustedMovement setting is enabled by default.

For testing, I changed the default from:

_unadjustedMovement = true;

to:

_unadjustedMovement = false;

After refreshing the Selkies web client, I verified in browser that Pointer Lock was being requested without the raw-input option:

requestPointerLock();

rather than:

requestPointerLock({
    unadjustedMovement: true
});

The sluggish/heavy mouse behaviour in Selkies Game Mode was dramatically improved.

Why this may be problematic as a default

unadjustedMovement: true requests raw mouse movement and bypasses operating-system mouse acceleration.

That behaviour can be useful when an application specifically requires raw mouse deltas, such as FPS camera movement.

However, standard Pointer Lock already provides relative mouse movement while retaining the platform's normal pointer adjustment.

For a remote application/game streaming system such as Selkies, raw input is therefore not necessarily the best default for every application.

For example:

FPS / first-person camera
    → raw input may be desirable

Strategy game / desktop-style cursor / older game
    → normal OS-adjusted Pointer Lock may feel substantially better

On my macOS client, the difference is large enough that enabling raw Pointer Lock makes Game Mode feel like a significant input-quality regression.

Expected behaviour

Activating Game Mode — fullscreen with pointer and keyboard locked should provide usable relative mouse input without unnecessarily changing the expected mouse response of the client platform.

Users who specifically require raw mouse input should still be able to enable it.

Actual behaviour

Current behaviour:

requestPointerLock({
    unadjustedMovement: true
});

On macOS:

sluggish / heavy / unnatural

Using:

requestPointerLock();

On the same macOS:

substantially smoother and more natural

Suggested solution

I don't think raw input needs to be removed from Game Mode.

Instead, unadjustedMovement could be treated as an optional input mode rather than always being enabled by default.

Possible approaches:

Option 1 — make raw input configurable

Expose a Game Mode/client setting such as:

Raw mouse input
[ ] Enabled

Users running FPS-style applications could explicitly enable it.

Option 2 — default to standard Pointer Lock

Use:

element.requestPointerLock();

by default, while allowing raw input to be enabled when required.

Option 3 — platform-aware default

If changing the global default would affect existing users, Selkies could at minimum avoid enabling unadjustedMovement by default for macOS clients.

A user-configurable option would still be preferable.

Client environment
Client OS: macOS
Browser: Helium (Chromium-based)
Input device: physical mouse
Selkies mode: Game Mode — fullscreen with pointer and keyboard locked

The original server-side test environment included Selkies, Pixelflux, Wayland, labwc, Wine, and a Windows game.

However, none of those server-side components are required to reproduce the difference between standard and raw Pointer Lock.

The behaviour can be reproduced locally using only browser's Pointer Lock API on macOS.

Additional context

There is an older Selkies issue concerning erratic/slow mouse movement in games (#28).

This report is specifically about the use of:

requestPointerLock({
    unadjustedMovement: true
});

when using Game Mode — fullscreen with pointer and keyboard locked, and the substantial behavioural difference this produces on my macOS/browser.

The isolated A/B result is:

macOS + Browser

requestPointerLock()
    → natural

requestPointerLock({ unadjustedMovement: true })
    → sluggish / heavy

The same improvement occurs in Selkies after changing _unadjustedMovement to false.

This suggests that raw mouse input should probably be configurable or opt-in for Game Mode rather than assuming that unadjustedMovement provides the best pointer behaviour for every client and application.

junkerderprovinz:
Closes #328.

`cee0b827` asks for `unadjustedMovement` on every platform. Linux and Android refuse it, so nothing changed there; macOS grants it, and what it removes is the curve that carried a slow hand across the remote screen. The client scales a locked delta onto the stream's pixels and quantizes it with a carry, but that gain is the same at every speed, so with the curve gone the pointer travels a fixed distance per count and feels heavy, which is what @doriguetto isolated down to two lines of plain HTML.

So the option is not asked for on macOS. The decision goes through the `browser` probe the file already keeps rather than a second one: `navigator.platform` decides and the client hint is only a fallback, since the question is which curve the OS applies, and an iPad in its desktop-class default reports `MacIntel` like a Mac and is told apart by its touch points. Windows keeps the option.

The audit covers the two cases a platform test gets wrong: an iPad, and a Mac whose client hint says nothing useful. `tests/tools/pointer_motion_page.html` now locks through the client's own request, so a run of `test_pointer_motion.py` measures the deltas that are really sent instead of a path the client no longer takes.

Whether one client-side constant would restore macOS properly is the open half, and it needs the measurement asked for in the issue. I have no Mac here, so someone on macOS has to confirm the feel before this is called done.


ehfd:
The one thing here that no macOS report can cover is the rig change: `pointer_motion_page.html` now locks through `_requestPointerLock` instead of asking for raw movement itself. Ran it on the platform where the option is refused, to check that path still measures what it did before.

`tests/e2e/test_pointer_motion.py` on Xvfb, 24/24 — Chrome and Firefox at dpr 1, 1.25 and 1.5, every one reporting `lockPath: plain`, and all three motion patterns landing within a pixel (40/40, 240/240, 480/480; a single 41 at dpr 1.5). The refusal path still ends in a lock, and the deltas still come out of the client's own relative-motion path rather than a model of it.

`tests/unit/test_pointer_lock.py` 37/37, the eight new platform cases included. The iPad-reports-`MacIntel` and client-hint-says-`Unknown` cases are the two a naive platform test gets wrong, and both hold.

The `isMacDesktop` comment also checks out rather than just reading well: `selkies-wr-core.js:426` and `selkies-ws-core.js:540` do tell an iPad from a Mac by `navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1`, so this really is the same test and not a second one.

ehfd (approved):

Issue #335, opened 2026-09-07
Udev interposer gets in the way of gpumanager in kwin 6.7 from finding a GPU and disables accell


thelamer:
I disabled the gamepad interposer on arch fedora and alpine because of this. 

Probably worth another visit. You can just scope apps with support but just in general our interposers should be non obtrusive to normal program operation and just be there for gamepads.

thelamer:
https://github.com/selkies-project/selkies/pull/336

PR #336, opened 2026-09-07
#335 rework entire fake udev concept

thelamer:
The old fake udev was written to respond with 0 or null when anything outside of gamepads was checked this mostly worked but some userspace stuff in our stack needs these, we ran into this with kwin 6.7 and their gpumanager asking libudev for drm devices. This refactor changes it to be distro specific based on the real libudev and passing the calls to it when they are requested. 

Tested across multiple problem containers Alpine 3.24, Arch, and Fedora 44. 

This beefs up the fake udev lib to not get in the way of normal program operation while still providing a working gamepad interposer backend for device discovery used by userspace programs.


thelamer:
Also fixed chromium and electron gamepad input while in here.

ehfd:
Validated this on hardware that actually has the devices the report is about — a 2×Tesla P100 box with `/dev/dri/{card0,card1,card2,renderD128,renderD129}` — since a container without render nodes cannot show the regression either way.

`drm` enumeration is identical with and without the preload:

```
without LD_PRELOAD:  6 devices  (card0, card1, card2, renderD128, renderD129, card0-VGA-1)
with    LD_PRELOAD:  6 devices  (same list, same order)
JS_LOG=1:            passing through to real libudev /usr/lib/x86_64-linux-gnu/libudev.so.1.7.12
```

So KWin's `gpumanager` gets its render nodes back, which is #335. `input` is unchanged too — the host's own `event0` and `mice` stay listed — until a pad socket is bound, at which point the pads appear beside them and nothing real disappears.

Also exercised:

- **musl/eudev** (Alpine 3.22): passthrough resolves `/usr/lib/libudev.so.1`, `drm` and `mem` match the unpreloaded scan, pads enumerate.
- **32-bit**: `make all32` builds clean, the `libudev_x86` path `addons/base/Dockerfile` uses.
- **`-Wall -Wextra`** clean on `fake-libudev-core.c` under both gcc and clang.
- `tests/unit/test_fake_udev_enumeration.py` 8/8; `tests/integration/test_gamepad_enumeration.py` 20/20 with one skip (no SDL2 headers here). The new fortified-open checks pass, `__read_chk` on the joydev init burst and `__open_2` returning the interposer's `EIO` for an unserved slot rather than the kernel's answer for that path included.

No blockers from me. One optional hardening, not worth holding this for: `initialize_virtual_gamepads_data_if_needed()` guards on a plain `bool` while `fake_udev_load_real` uses `pthread_once`. In practice `udev_new` runs the init first and single-threaded, so the only window is two threads racing the very first `udev_new` — a second `pthread_once` would close it for free.

ehfd:
Hardened this rather than leaving it as a note — pushed `f4d01e5d3`. It turned out not to be theoretical.

Both lazy inits are now `pthread_once`, matching what `fake_udev_load_real` already did:

- `initialize_virtual_gamepads_data_if_needed()` splits into the builder plus a `pthread_once` wrapper, so the name the module comment references still exists.
- `fake_udev_logging_init_if_needed()` was the worse of the two: it stored `initialized = true` **before** reading `JS_LOG`, so a second thread could see the flag published and start logging against a `g_fake_udev_log_enabled` that had not been written yet.

Measured with ThreadSanitizer — eight threads each building their first context and scanning `input`, released from a barrier so they arrive together:

```
before:   81 data races
            'initialized.9'     in fake_udev_logging_init_if_needed
            'virtual_gamepads'  (7712 bytes) — concurrent snprintf into the shared table
after:     0 data races
```

The `virtual_gamepads` one is the reachable consequence: a thread can observe the table as initialized while another is still `snprintf`-ing into `input_phys` / `input_uniq` / `usb_serials`, and hand out a half-written syspath.

Re-verified after the change: unit 8/8, integration 20/20 (+1 skip), gcc and clang `-Wall -Wextra` clean, 32-bit `all32` builds, and on the P100 box `drm` still enumerates identically with and without the preload (6 = 6) with the pads appearing on `input` — under Alpine/musl-eudev as well as glibc.

I left the interposers alone deliberately. Their `if (!real_x && load_real_func(...))` fallbacks are unsynchronized too, but they memoize an idempotent `dlsym` result into an aligned pointer that the constructor has already resolved — that is not shared state being constructed, so converting ~40 of them would be churn with no behavioural change. Their `getenv` reads all happen inside the constructor, single-threaded at load.

ehfd (approved):

PR #350, opened 2026-09-11
Take a release's pinned capture stack from the projects' own releases

ehfd:
A release resolved the pinned pixelflux and pcmflux from PyPI, which is where they are not: PyPI carries 2.0.0 of
each while this repository pins `~=2.1.0`, so the smoke test, the images, the packages and the AppImages of a
release run would each fail in pip's resolver. The capture-stack workflow already knows how to take wheels from
those projects' releases, which it does per commit for every build ahead of a release, so a release now does the
same by version: `locate` reads the pins out of `pyproject.toml`, names the release each project cut for that
version, and `fetch` downloads it into the run's artifacts, where everything downstream already reads them. A
version no release carries is built from its tag, as a main HEAD with no pre-release already was.
`capture_stack: pypi` keeps the old path, for a release whose capture stack is published there.

**Reproduced before the change.** `locate`, lifted out of the YAML and run against the real repositories: for
`main` it names each project's per-commit pre-release (`a3290fd`, `a489f79`) and fetches it, and there was no
path at all that could name a version — a release only ever set `capture_stack_published: true`, and
`pip install pixelflux~=2.1.0` has nothing to resolve, since neither project has ever cut a version-tagged
release and PyPI stops at 2.0.0.

**Gone on this code.** The same step, run against the real repositories for the HEAD path and against a stubbed
`gh` for the release shapes that do not exist upstream yet; 17 checks green, and the hermetic half of them is now
`tests/unit/test_capture_stack.py` (20 checks), which also holds `pyproject.toml`'s pins to a spelling the step
can read:

| what `locate` is given | what it does |
|---|---|
| `main`, a per-commit pre-release for its HEAD | fetches that pre-release |
| `main`, no pre-release for its HEAD | builds `main` |
| a pin, a release carrying wheels of that version | fetches that release |
| a pin, a release whose wheels are of another version | builds `refs/tags/<pin>` |
| a pin, no release at all | builds `refs/tags/<pin>` |
| no pin in `pyproject.toml` | fails the run |

`~=`, `==`, `>=` and a `.dev` pin all read back as the version they name. The fetch itself is the same
`gh release download --pattern` a HEAD run uses, measured against a real release: 7 wheels, cp39 through cp315,
for one (arch, libc) slice.

**Reach.** The wheels workflow gains an input and two outputs per project; `release.yaml` gains the job and the
choice; `images.yaml`, `packages.yaml` and `smoke-wheel.yaml` only lose "main HEAD" from messages that now
describe either source. The draft release drops the capture-stack wheels before it attaches anything, so a release
carries what it carried before. `pre-commit` (ruff, codespell), `actionlint` 1.7.12 with the pinned shellcheck,
and the whole unit tier (87 suites) are green; the job graph itself is static-checked only, since running it
means cutting a release.

One thing a maintainer has to spell: pip takes no pre-release for a pin that does not name one, so a release
candidate capture stack is pinned as `pixelflux~=2.2.0rc1`. Measured both ways — `~=2.2.0rc1` installs the rc,
`~=2.2.0` refuses it.


ehfd:
The pinned path was dispatched on this branch to see it on a runner rather than only in the harness:
[run 34602180905](https://github.com/selkies-project/selkies/actions/runs/34602180905). `locate` read both pins out
of `pyproject.toml`, found no release carrying them, and set `build`, `pixelflux_ref` and `pcmflux_ref`:

```
pixelflux: nothing released for refs/tags/2.1.0; building it
pcmflux: nothing released for refs/tags/2.1.0; building it
```

The build jobs then failed where they should, by name:

```
ref: refs/tags/2.1.0
fatal: couldn't find remote ref refs/tags/2.1.0
```

That is the run deliberately left red: it is what a release cut before its capture stack exists looks like, and
it is the state today — neither project has ever cut a version-tagged release, and PyPI stops at 2.0.0 while this
repository pins 2.1.0. Tag `2.1.0` in pixelflux and pcmflux (selkies-project/pixelflux#15 and
selkies-project/pcmflux#2 make that tag the version its wheels carry) and the same dispatch fetches them instead.

PR #353, opened 2026-09-12
feat(clipboard): Carry formatted content, and switch the clipboard and chords from the panel

ehfd:
Three things that share the clipboard and input paths.

**Formatted content keeps its formatting.** A copy that carries markup now travels with the plain text its source wrote for it, in one envelope, so pasting into a rich editor keeps the styling and pasting into a terminal gets the text rather than a rendering of the markup. The X11 selection owner holds a set of flavours and answers each conversion request with the one it names; the Wayland paths do the same. Copy-out prefers a picture over the markup wrapped around it, and markup over the plain text beneath it, so copying an image still yields the image. Markup is gated as text, not as an image, so it does not need the image clipboard on.

**Clipboard switches in the panel.** Send to session, Receive from session and Seamless, in both dashboards. The first two narrow whatever `--enable-clipboard` already allows, and a direction the server refuses is not offered at all. Seamless is what makes the clipboard follow every copy by itself; turned off, the panel's clipboard box becomes the deliberate path in both directions. `--clipboard-seamless` sets the starting position and can be locked.

**The client's chords can be handed to the session.** Control+Shift with F, M, X or G, and Control+Shift+click, are taken by the client before the session sees them, which collides with an application inside the session binding the same ones. The shortcuts section now has a switch that passes them all through; the panel's buttons still reach every function and the triple-Escape hatch still leaves gaming mode. `--keyboard-shortcuts` sets the starting position and can be locked.

Verified, not assumed. An independent X client converting the selection target by target gets the markup from `text/html` and the plain text from `UTF8_STRING`, and `wl-paste` gets the same from a real headless compositor. In Chromium against a live session, on both transports, a rich copy crosses in both directions with both flavours intact. With seamless off a copy on either side leaves the other alone; with the chords handed over, Control+Shift+F stops taking the browser and arrives in the session instead, which the X server's own keymap confirms. New suites in the unit, integration and end-to-end tiers, plus the existing clipboard suites and the unit tier.

Needs selkies-project/pixelflux#16, which makes the compositor serve one payload per flavour instead of one payload under several names.


ehfd:
This carries the two discussions it answers, since a pull request cannot close a discussion on its own.

- "How to disable selkies shortcuts?" (#201): the shortcuts section's switch hands Control+Shift+F and the other chords to the session, so an application that binds them keeps them. Gaming mode already passed keystrokes through by locking the keyboard, but it did not stop the client taking those four chords; this does, without giving up the chords for anyone who wants them.
- "Toggle clipboard synchronization options in panel" (#221): the clipboard section gained the two directions and the seamless switch, so the sync can be narrowed or made deliberate per browser without changing the server's policy.

The rich clipboard work rides along because it touches the same paths: the flavour set had to exist before the panel could talk about what moves.

PR #359, opened 2026-09-13
pull back the shift neutralization a bit for function key handling

thelamer:
This got a heavy hand to fix some input translation bugs but caused a regression from the previous input stack. 

Things like shift+Home or shift +End are supposed to select text to line end or beginning, same with arrow navigation. 

This covers the known function keys that respond to shift modifiers and adds a unit test to ensure we do not break it in the future.


junkerderprovinz:
🔔 repo-watch: shipped in 2.0.0rc1.

PR #397, opened 2026-09-17
feat(capture): Ask the browser for the camera and microphone only while the session uses them

gutschke:
## Why

A session configured for the camera or microphone asks the browser for it at startup and holds it until the session ends, so the recording indicator on the user's own hardware is lit for the whole session whether or not anything on the remote desktop uses the device. This makes the ask follow the use.

## What

`--webcam-on-start` and `--microphone-on-start` take a third value beside `true` and `false`: `demand`. The server then watches what reads the virtual device (the interposer's clients, PipeWire's consumers and a kernel device's openers for the camera; the sound server's uncorked recording streams for the microphone) and asks one page to capture only while something does, releasing it after a hold-off (2 s for the camera, 10 s for the microphone, each needing three idle polls). Only one page is ever asked, since every page that answers pushes into the same device: the page asked last is told to release before another is asked or once it may no longer capture (closed, demoted, moved to a secondary display). A shared viewer is never a candidate, a `*_enabled` locked off still wins, a toggle made during the session outranks the demand, and a page that refused is not asked again until the user toggles. Every transition is logged with the reader's name and audited as `capture.demand`.

The default stays `false`; `demand` is opt-in because any process on the desktop can then cause a permission prompt, and under it the camera device exists before any uplink, so a `webcam_pixel_format` of `auto` resolves to I420.

## Where

- `src/selkies/capture_demand.py`: the watch, the hold-off and the routing, once for both transports; each transport contributes `capture_candidates()` and `tell_capture()` and calls `capture_demand.sync()` where its candidate list changes.
- `webcam.py`: `VirtualWebcam.consumers()` serves the format decision and the demand watch (`assume_reader`), with the /proc walk read through `readlink` and held to one per four seconds on the demand path.
- `audio_control.py`: `PulseNode.corked` and `AudioControl.recorders()`.
- Both cores: `CAPTURE_DEMAND <subject> <0|1>` / `capture_demand,<subject>,<0|1>`, obeyed unless the user toggled the device; a refusal latches (`isCaptureRefusal`).
- Suites: `unit/test_capture_demand.py`, `unit/test_capture_demand_readers.py`, `integration/test_capture_demand.py`, `e2e/test_webcam.py ondemand`.


ehfd:
@gutschke Thanks.

I get the point, but I need to diet massively on the code volume. Please wait until I come up with more commits.

ehfd:
Thanks for this, the feature is wanted. I reworked the branch in place (maintainer edits are on) so the production footprint matches its size, and force-pushed it as one commit with you as the author and me as co-author. What changed and why:

**Settings.** No new flags. `webcam_on_start` and `microphone_on_start` are now three-valued strings like `enable_clipboard`: `true`, `false` (the default, unchanged behavior) and `demand`. The client keeps reading booleans; `build_client_settings_payload` derives `*_on_start` and `*_on_demand` from the policy, the way the clipboard gate booleans are derived. `webcam_unanswered_timeout` is gone (below).

**One watcher, one router.** `capture_demand.py` holds the poll loop, the hold-off and the routing. Each transport contributes `capture_candidates()` and `tell_capture()` and calls `capture_demand.sync(self)` where its candidate list changes, so the "page told last is released before another is asked" rule exists once instead of once per transport, and a send that fails is retried at the next poll rather than lost. `mic_demand.py` folded into the same module; `WebcamDemand.reader` reuses `VirtualWebcam.consumers()` (the former `_consumers`) with an `assume_reader=False` argument instead of a second predicate, and the /proc walk floor lives in that one place.

**Dropped: the unanswered-reader deadline.** Stopping the process-wide camera to unblock one reader, with exponential backoff and a re-create on the next poll, was the bulk of the fragile code (your own note lists seven bisect-breaking commits for it). The blocked `VIDIOC_DQBUF` it addressed predates this feature: on `main` a camera once used outlives its uplink for the process's life and a later opener blocks the same way. If a bound is wanted it belongs in the interposer (which already ends readers with `ENODEV` when the server goes), not in a per-frame counter on the server.

**Kept from your series:** the wire verb (`CAPTURE_DEMAND <subject> <0|1>` / `capture_demand,<subject>,<0|1>`), the `corked` field and `recorders()`, the readlink walk, the refusal latch (`isCaptureRefusal`, one predicate), the integration suite and the browser block, the audit event (now a documented `capture.demand` row).

**Numbers on this branch** (sandbox: rustbuild env, pixelflux built from its `main`, private Xvfb and PulseAudio):

| | |
|---|---|
| production lines (`src` + `addons`) | +466/−51 in 11 files, was +1018/−21 in 12 |
| whole diff | +1259/−67, was +2187/−24 |
| unit tier | 117 suites pass (`test_capture_demand` 40 checks, `test_capture_demand_readers` 39) |
| `integration/test_capture_demand.py` | 5/5, microphone half included |
| `e2e/test_webcam.py ondemand` | 28/28 over websockets and webrtc on chromium, firefox and webkit; 1 skip (firefox webrtc: no H.264 in the Playwright build, as `transport_block` skips) |
| `/proc` walk, 309 processes / 11 668 descriptors | 70 ms with `readlink`, 460 ms with `realpath` |
| idle server, one page connected, both watches | 0.92% of a core vs 0.76% without |
| open → page asked → 30 frames at the device | chromium 1.0 s / 4.6 s, firefox 0.35 s / 1.9 s, webkit asked at 0.14 s (frames: see below) |

**WebKit.** Your block ran WebKit against the status flips only, and it turns out that is all it can prove: Playwright's WebKit never opens the interposer camera (the published camera's client count stays 0 while WebKit "captures"), it uses its own mock device, and that device delivers frames to a `<video>` element only about half the time headless (10 of 10 stalls on a bare page, 3 of 6 on the dashboard, the same under the side-menu toggle without any demand). So the block now proves frames at the device on Chromium and Firefox, which capture the published camera, and the ask/release routing on all three engines. pixelflux itself delivers JPEG to a reader that connected before the first frame 8 of 8 times, so nothing in the server path is implicated.

Two things I could not run: the audit webhook end to end (checked by unit fakes only) and a real v4l2loopback node for the direct-open path (no module in this sandbox; the walk itself is covered against a child process holding a file). Firefox over WebRTC is skipped for the reason the existing block gives.

gutschke:
No objection to the rework; in fact, impressed with such a fast response :-) Thanks

Dropping the unanswered-reader deadline was the part I was least confident in. Very pleased with deferring to your perspective as a maintainer.

For the record only, not suggesting a change here: str(...).split("|")[0] drops a |locked suffix on the webcam_on_start and microphone_on_start settings. It's the same line enable_clipboard already uses. I assumed it as house convention rather than something you introduced.

ehfd:
Follow-up verification after the push, closing the three items the first comment listed as not run, plus two more:

| Check | Result |
|---|---|
| audit webhook end to end (local collector) | `capture.demand` `ask` with `reader=interposer client` at +1.0 s, `release` at +9.0 s, beside the `session.*` events |
| real v4l2loopback node (P100 host, `v4l2loopback-dkms`, `/dev/video42`, base image + this branch) | the server mirrors into the node; a plain `open()` with no interposer has the page asked within the 4 s walk floor (`an application holding /dev/video42`) and released after close; the /proc walk costs 0.8 ms in the container |
| Firefox over WebRTC (OpenH264 seeded in the e2e profile) | 30 frames 1.5 s after the open; the browser block is 34/34 with no skips |
| live transport switch (`/api/switch` WebSockets → WebRTC in one process) | 12/12: asked, served and released on both sides, one camera for the whole run |
| two WebSocket controllers, the asked one disconnects | the other is asked in its place and released when the device closes |
| two WebRTC controllers | the signaling server evicts the first for the second (same identity), which exercises reap-then-ask; a held camera handed between two token-separated WebRTC controllers is covered by the unit routing only |

Also amended: the start-state invariant in `AGENTS.md` names the `demand` policy. The seven failures `integration/test_webcam_device.py` shows in my sandbox are its PipeWire checks and are identical on `main` (the daemon there does not answer `pw-cli`).

thelamer:
Passed my smoke tests on android, iOS, and desktop at latest commit.

ehfd:
@gutschke Right, the suffix is dropped the way `enable_clipboard` drops it; the settings reference scopes `|locked` to bool settings, and on these two it never had a consumer (`pipeline_starts_on` and both cores read the value alone). Your note did surface one thing worth a final amendment: the bool rule also accepts `1`, so a deployment with `SELKIES_WEBCAM_ON_START=1` would have fallen back to `false` with a warning. The head now reads `1`/`0` as `true`/`false` for both policies (covered in `test_media_on_start`). Nothing else changed; CI is green on the previous head and queued on this one.

@thelamer Thanks for the smoke run.

ehfd (approved):

PR #398, opened 2026-09-17
Move away from xrandr for multi monitor x11 and use real monitors

thelamer:
So xlibre supports crtcs which means we could technically boot it with like --crtcs 4 but it makes more sense to patch it to allow plugging and unplugging physical monitors in the software stack. The net result is we call on that instead of xrandr for monitor creation with a virtual split. The split is real now and honored by all child desktop environments as far as they are concerned these are real monitors and are treated as such. 

We also had a bug where we were not giving primary movement enough time for DEs so if you open a monitor leftward previously a bunch of child programs like plasmashell move the primary X units right to which is correct, but when we destroy this leftward monitor there was not a specific primary move back to 0,0 so some programs could render off the primary display, those are the only real logic changes in selkies outside of the new crtc path for diplay utils. 

To keep things clean the now legacy xrandr methods are shifted into their own file and called only as a fallback if xvfb does not support the protocol we patch in for dynamic creation and deletion of monitors. 

The beauty of this approach is it is DE agnostic, if they support multiple monitors and are X11 based they work. 

Certified mate, xfce, kde, lxqt all working now.


ehfd:
Reviewed with the patched Xvfb built here (XLibre 25.2.2 + both patches) and on the P100 node with the published KDE and LXQt images running this branch, then squashed to one commit and force-pushed. The suites job already passed on the identical content with the Xvfb the runner now builds (run 35427438460, 171 suites, no skips), and the rerun on the squashed head is in progress.

**What was measured**

- On the patched server, GTK 3 and Qt 5 each see two monitors (`screen`, `screen_1`); a maximized window stays on one output under Openbox (no restart) and KWin 5.27. With logical monitors on the same server, Qt sees one 2304-wide screen and KWin maximizes across both, which is the defect this fixes.
- Plasma 6 (glx-desktop image, Xvfb handover) and LXQt (desktop image): a second desktop window appears per output for right and left placements, the panel follows the primary's move, and both go away when the page leaves. Both transports pass the two-display, cross-display drag, two-display pixels and locked-cursor suites on the patched display.
- The staged move for a left/top display reaches Qt as a geometry change before the added screen. Plasma 6 laid itself out correctly without the stage in three runs as well, so I could not reproduce the need for it; it is kept.
- Stock `xrandr` can drive the property too (`--set Connected 1`, then a mode).

**Changed on top**

- A display keeps the output it holds from one layout to the next. Before, outputs were handed out in display order, so with three displays the middle one leaving re-plugged the last one: Qt saw `removed:screen_2` while the display was still there. The stage is computed from the live displays rather than a count for the same reason.
- The layout skips CRTC and screen size settings already in place. Every Xvfb (stock, pristine XLibre, patched) blanks its framebuffer on any `RRSetScreenSize`, the same size included, so a reconfigure that changed nothing left the desktop black until windows repainted; the locked-cursor suite caught it (7/12, now 12/12).
- `addons/base/build-xvfb.sh` is one recipe for the base image and the tests workflow, cached like labwc, so the suites drive the server the images ship; `test_output_layout` no longer skips in CI. The suites that keyed on `selkies-*` monitor names (`test_mixed_dpi`, `test_switch_stress`, the two-display WM restart check) now read the primary flag or the outputs, and the constrained-root suite falls back to the distribution's Xvfb because the patched one never caps its root.
- `docs/component.md` describes the output layout ahead of the logical-monitor paragraph.

**Not this PR**

- `test_mixed_dpi` "primary x1, secondary x2" and two switch-stress checks fail identically on main here (the first layout waits on a 10 s `pactl` stall in this sandbox).
- LXQt showing one wallpaper stretched over both displays is `PerScreenWallpaper` not being in effect: the published desktop image predates the default main sets, and pcmanfm-qt writes its user settings file on exit, so a home that ran an older image keeps `false`. With the value in effect both paths paint each screen its own wallpaper.
- XLibre initializes its CRTC size before `-screen` is parsed (a `-screen 0 1280x720x24` server reports a 1280x1024 mode until the first mode set); the images pass no `-screen`.

